PR #33, opened 2014-08-16
First pass at bitmap mask support

zewt:
This is functional but shouldn't be merged as-is.  I'm sending a PR because it's been sitting around in a "good enough for now" state for a while and I want to make it available if anyone wants to work on it (or use it as-is), rather than having it be lost on my drive.

This implements bitmap masks, which applies a layer's mask to the bitmap when it's converted to an image.  It doesn't provide any interface to read or manipulate the mask directly, but it's enough to handle the common case where you just want to get the flattened layer pixels.  Note that the mask enabled flag is ignored; masks are always applied.

This only looks at ChannelID.USER_LAYER_MASK.  I don't know what REAL_USER_LAYER_MASK is, so I left it unhandled.

This is missing:
- extract_composite_image support.  The data flow here is different from extract_layer_image, and I haven't looked at what to do here.
- There are no tests.

### DIFF
--- a/src/psd_tools/user_api/pil_support.py
+++ b/src/psd_tools/user_api/pil_support.py
@@ -7,7 +7,7 @@ from psd_tools.constants import Compression, ChannelID, ColorMode, ImageResource
 from psd_tools import icc_profiles
 
 try:
-    from PIL import Image
+    from PIL import Image, ImageChops
     if hasattr(Image, 'frombytes'):
         frombytes = Image.frombytes
     else:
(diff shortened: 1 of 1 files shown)


flux627:
I'm looking to contribute to this PR- I need to be able to export a layer's mask data as a PIL object. Any hints to get my started in the right direction would be appreciated. Also, advice as to how the best way to contribute here? Pull request to the fork? Grant me write access to the fork? New PR?

PR #39, opened 2015-01-26
extraction of embedded files

vbraun:
Access for embedded files, these are called embedded smart objects by Adobe I think.

I'm using this to save embbedded Illustrator files as SVG without using any Adobe product using

    for i, emb in enumerate(psd.embedded):
        emb.save_as_svg('/tmp/psdexport/icon-{0}.svg'.format(i))

### DIFF
--- a/src/psd_tools/decoder/linked_layer.py
+++ b/src/psd_tools/decoder/linked_layer.py
@@ -1,0 +1,80 @@
+# -*- coding: utf-8 -*-
+from __future__ import absolute_import, unicode_literals, print_function
+import warnings
+import io
+import struct
+
+from psd_tools.utils import read_fmt, read_pascal_string, read_unicode_string
+from psd_tools.debug import pretty_namedtuple
+
+LinkedLayerCollection = pretty_namedtuple('LinkedLayerCollection', 'linked_list ')
+_LinkedLayer = pretty_namedtuple('LinkedLayer',
+                               'version unique_id filename filetype creator decoded uuid')
+
+
+class LinkedLayer(_LinkedLayer):
+
+    def __repr__(self):
+        return "LinkedLayer(filename='%s', size=%s)" % (self.filename, len(self.decoded))
+
+    def _repr_pretty_(self, p, cycle):
+        if cycle:
+            p.text("LinkedLayer(...)")
+        else:
+            with p.group(1, "LinkedLayer(", ")"):
+                p.breakable()
+                p.text("filename='%s', ", self.filename)
+                p.breakable()
+                p.text("size=%s, ", len(self.decoded))
+                p.breakable()
+                p.text("unique_id=%s, ", self.unique_id)
+                p.breakable()
+                p.text("type='%s', ", self.filetype)
+                p.breakable()
+                p.text("creator='%s', ", self.creator)
+
+
+def decode(data):
+    """
+    Reads and decodes info about linked layers.
+
+    These are embedded files (embedded smart objects). But Adobe calls
+    them "linked layers", so we'll follow that nomenclature. Note that
+    non-embedded smart objects are not included here.
+    """
+    fp = io.BytesIO(data)
+    position = 0
+    layers = []
+    while True:
+        length_buf = fp.read(8)
+        if not length_buf:
+            break   # end of file
+        length = struct.unpack('>Q', length_buf)[0]
+        liFD, version = read_fmt('4s I', fp)
+        if liFD != b'liFD':
+            warnings.warn('unknown layer type')
+            break
+        unique_id = read_pascal_string(fp, 'ascii')
+        filename = read_unicode_string(fp)
+        filetype, creator, remaining_length, file_open_descriptor = read_fmt('4s 4s Q B', fp)
+        filetype = str(filetype)
+        if not file_open_descriptor:
+            if remaining_length + 198 != length:
+                warnings.warn('record length mismatch')
+        else:
+            # WTF is this, Adobe: "variable length: Descriptor of open parameters"
+            warnings.warn('decoding of file open descriptor not implemented')
+            size = length - remaining_length - 198 + 4   # undocumented guess
+            fp.read(size)
+        decoded = fp.read(remaining_length)
+        uuid = read_unicode_string(fp)
+        layers.append(
+            LinkedLayer(version, unique_id, filename, filetype, creator, decoded, uuid)
+        )
+        # Each layer is padded to start at 4-byte boundary
+        position += length
+        pad = -position % 4
+        fp.read(pad)
+        position += pad
+    return LinkedLayerCollection(layers)
+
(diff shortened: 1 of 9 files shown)


kmike:
Hi @vbraun,

Thanks - the code looks good!

Could you please check the test faiures? It seems this PR breaks parsing of some PSD files (try smart-object-slice.psd file). 

If these faiures are hard to fix or if you expect more of them then it is better to stop earlier in the new `decode` function instead of showing a warning, in order to prevent an expection in later code. This will allow to extract at least some information in cases the existing decoder of embedded objects doesn't work.

src/psd_tools/user_api/embedded.py
```diff
@@ -0,0 +1,83 @@
...
+            self.save(tmp_file)
+            yield tmp_file
+        finally:
+            if os.path.exists(tmp_file):
+                os.remove(tmp_file)
+            os.rmdir(tmp_dir)
+
+    def save_as_png(self, filename=None):
+        """
+        Save embedded file as SVG
```
kmike: "as PNG"

src/psd_tools/decoder/linked_layer.py
```diff
@@ -0,0 +1,82 @@
...
+    non-embedded smart objects are not included here.
+    """
+    fp = io.BytesIO(data)
+    layers = []
+    while True:
+        start = fp.tell()
+        length_buf = fp.read(8)
+        if not length_buf:
+            break   # end of file
+        length = struct.unpack('>Q', length_buf)[0]
```
kmike: str('>Q') should fix test failures

src/psd_tools/decoder/linked_layer.py
```diff
@@ -0,0 +1,82 @@
...
+        liFD, version = read_fmt('4s I', fp)
+        if liFD != b'liFD':
+            warnings.warn('unknown layer type')
+            break
+        unique_id = read_pascal_string(fp, 'ascii')
+        filename = read_unicode_string(fp)
+        filetype, creator, filelength, file_open_descriptor = read_fmt('4s 4s Q B', fp)
+        filetype = str(filetype)
+        if file_open_descriptor:
+            # WTF is this, Adobe: "variable length: Descriptor of open parameters"
```
kmike: A guess: maybe this data can be decoded as a Descriptor (see https://github.com/kmike/psd-tools/blob/master/src/psd_tools/decoder/actions.py#L84)?

vbraun:
It seems related, but is not quite what `decode_descriptor` understands. It doesn't start with a unicode string, e..g.  In placedLayer.psd the mysterious "file open descriptor" is

    '\x00\x00\x00\x10\x00\x00\x00\x01\x00\x00\x00\x00\x00\x00null\x00\x00\x00\x01\x00\x00\x00\x08compInfoObjc\x00\x00\x00\x01\x00\x00\x00\x00\x00\x00null\x00\x00\x00\x02\x00\x00\x00\x06compIDlong\xff\xff\xff\xff\x00\x00\x00\x0eoriginalCompIDlong\xff\xff\xff\xff'

kmike:
Yeah, that's something different indeed. I don't understand what is going on before `\x08compInfoObjc`. `\x08compInfoObjc` and beyond looks like the data decoded by this code: https://github.com/kmike/psd-tools/blob/2a73b59fb72ba7ba35d75267cb7f250be08a71cb/src/psd_tools/decoder/actions.py#L92

kmike:
`\x00\x00\x00\x01` could be "item count", but I'm not sure. "null" is classID. 4 zero bytes before it is classID_length.

vbraun:
Thats what I'm thinking, too:
~~~~~
\x00\x00\x00\x10   # some integer = 16

\x00\x00\x00\x01\x00\x00  # unicode empty string

\x00\x00\x00\x00   # classID length 0 means: 4
null               # classID

\x00\x00\x00\x01   # descriptor item count

\x00\x00\x00\x08   # key length
compInfo           # key
Objc               # ostype = descriptor

\x00\x00\x00\x01\x00\x00  # unicode empty string

\x00\x00\x00\x00   # classID length 0 means: 4
null               # classID

\x00\x00\x00\x02   # descriptor item count

\x00\x00\x00\x06   # key length
compID             # key
long               # ostype = integer
\xff\xff\xff\xff   # integer

\x00\x00\x00\x0e   # key length
originalCompID     # key
long               # ostype = integer
\xff\xff\xff\xff   # integer
~~~~~

kmike:
yeah, this works: `decode_descriptor(None, BytesIO(data[4:]))`

vbraun:
Its still a mystery to me why Adobe writes this in the file, but it does decode:

    Descriptor(name=u'', classID='null', items=[('compInfo', Descriptor(name=u'', classID='null', items=[('compID', Integer(value=4294967295)), ('originalCompID', Integer(value=4294967295))]))])

kmike:
Could you please preserve these extra fields (`have_file_open_descriptor`, `file_open_descriptor` and `uuid`, anything else?) in LinkedLayer data structure?

kmike:
Even if they don't look useful it is nice to be able to restore the original data from the decoded data as much as possible.

src/psd_tools/decoder/linked_layer.py
```diff
@@ -0,0 +1,82 @@
...
+        layers.append(
+            LinkedLayer(version, unique_id, filename, filetype, creator, decoded)
+        )
+        # Undocumented extra field
+        if version == 5:
+            uuid = read_unicode_string(fp)
+        # Gobble up anything that we don't know how to decode
+        expected_position = start + 8 + length      # first 8 bytes contained the length
+        if expected_position != fp.tell():
+            warnings.warn('skipping over undocumented additional fields')
```
kmike: Is this still needed? If it is needed, it'd be good to preserve this extra binary data in LinkedLayer.

vbraun: I don't have an example where it is needed, but in the interest of robust parsing I think its good to have.

vbraun:
Since `have_file_open_descriptor == (file_open_descriptor is not None)` I don't think we need to store it, agree?

kmike:
yes, it makes sense

src/psd_tools/user_api/embedded.py
```diff
@@ -0,0 +1,83 @@
...
+    def __init__(self, linked_layer):
+        self._layer = linked_layer
+        self.filename = linked_layer.filename.strip('\0x0')
+        self.unique_id = linked_layer.unique_id
+
+    def preferred_extension(self):
+        filetype = self._layer.filetype   # 4 chars: 'png ', 'PDF ', ...
+        return filetype.lower().strip()
+
+    def get(self):
```
kmike: What do you think about renaming it to `data()` or `@property data`?

vbraun:
I don't have a preference, whatever you think fits best

kmike:
The PR looks good :+1: 

Could you please add yourself to https://github.com/kmike/psd-tools/blob/master/AUTHORS.txt? I'll merge this PR after this last fix.

PR #41, opened 2015-03-03
Added layer.replace_image.

duanemalcolm:
Not sure if this is useful or the correct design. I added it so I could replace the screenshot in a phone mock-up.

This also my first pull-request. I'm not sure if I'm doing it right.

Cheers, Duane


duanemalcolm:
Maybe it should be called "image_data" to be consistent with the "text_data" property just above it.

duanemalcolm:
That was a lame attempt. Clearly the refactoring didn't work like I thought in PyCharm. I'll fix the bug and try again. Might change it to "image_data" too.

PR #43, opened 2015-03-10
monkey patch for PIL 1.1.7 for CMYK files

michaelmwu:
Version 1.1.7 of PIL seems to lack tobytes. I think someone messed up on the 2 to 3 conversion. Unfortunately I can't upgrade right away...

I didn't want to patch PIL globally so just doing it on the object


kmike:
Hi,

This change makes sense, but I think that it is better to solve this issue without patching PIL, by introducing a new function 'image_tobytes' (or simply 'tobytes') like it is done for 'frombytes' - see the top of the file.

michaelmwu:
since it's a method on an object, maybe we can just do logic to decide which method to call?

kmike:
yeah, this also makes sense

michaelmwu:
how's that?

kmike:
:+1:

kmike:
Could you please add yourself to AUTHORS.txt?

PR #49, opened 2015-08-26
Blend modes support for merging

EvgenKo423:
Hi, Mikhail! I implemented Photoshop blend modes for layer merging, though I can't implement fill opacity.


kmike:
Awesome! I have some comments, but first is what do you think about extracting blend modes extension to a separate library (blendosaurus? blend-modes?) We can put it under psd-tools organization.

EvgenKo423:
It's actually a good idea! But in that case I'd prefer to make my own repo. Also, I think this "library" is not quite ready for a stand-alone release, at least checks for fools should be added.

EvgenKo423:
I see a lot of progress has been done on layer composition, but I don't understand why the blend modes were re-written from scratch (bringing another dependency, which is quite heavy, especially with MKL support) when they were already implemented and fine-tuned. And I don't see any point in making floating point computations if the result is always an 8-bit image anyway...

So I just released my extension as a standalone module for everyone and then it's up to you whether you'll use it or not.

I also compared the `psd-tools` implementation (using `numpy` 1.16.6 with MKL support) against `image-blender` (the standalone module) and `psd-tools` was 13.5 times slower on average. The following table shows the differences in precision:

Function name |    Maximum deviation (image-blender)    | Maximum deviation (psd-tools)
--------------|-----------------------------------------|------------------------------
--------------|------------ **Normal group** -----------|------------------------------
dissolve      | Deviations approach normal distribution | Not implemented
--------------|------------ **Darken group** -----------|------------------------------
darken        | 0                                       | 0
multiply      | 0                                       | -1 color tone
color_burn    | 0                                       | -1 color tone
linear_burn   | 0                                       | -1 color tone
darker_color  | 0                                       | -1 luminance level
--------------|----------- **Lighten group** -----------|------------------------------
lighten       | 0                                       | 0
screen        | 0                                       | -1 color tone
color_dodge   | 0                                       | -1 color tone
linear_dodge  | 0                                       | 0
lighter_color | 0                                       | -1 luminance level
--------------|----------- **Contrast group** ----------|------------------------------
overlay       | 0                                       | -1 color tone
soft_light    | ±1 color tone                           | -1/+16 color tones
hard_light    | 0                                       | -1 color tone
vivid_light   | +2 color tones                          | ±127 color tones
linear_light  | 0                                       | -1 color tone
pin_light     | +1 color tone                           | +1 color tone
hard_mix      | 0                                       | -127 color tones
--------------|---------- **Inversion group** ----------|------------------------------
difference    | 0                                       | -1 color tone
exclusion     | ±1 color tone                           | -1 color tone
--------------|--------- **Cancellation group** --------|------------------------------
subtract      | 0                                       | -1 color tone
divide        | 0                                       | -1 color tone
--------------|---------- **Component group** ----------|------------------------------
hue           | ±2 luminance levels                     | -127/+128 color tones
saturation    | ±2 luminance levels                     | -127/+128 color tones
color         | ±2 luminance levels                     | -2/+1 luminance levels
luminosity    | ±2 luminance levels                     | -2/+1 luminance levels

kyamagu:
@EvgenKo423 Thanks for reporting that, the current implementation attempts to implement the transparency model of PDF spec. See Chapter 7 of the reference https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/pdf_reference_archives/PDFReference.pdf

There, apparently 8-bit image ops are not sufficient. Also check some of the testing fixtures in the repo

PR #54, opened 2015-12-05
Fix incorrect conditional when attempting to decode references.

joshdrake:
The interaction I ran into involved smart objects and smart filters, but I couldn't manage to reproduce the problem in a fresh PSD. I've uploaded a rather large failing PSD here as an example.

Failing PSD: [test_enum.psd.zip](https://github.com/psd-tools/psd-tools/files/52968/test_enum.psd.zip)


kmike:
Thanks @joshdrake! Is it OK to add your example PSD file to the testing suite?

joshdrake:
You bet, thanks for merging!

PR #67, opened 2017-01-02
namedtuple's verbose argument changed to keyword-only in Python 3.6

objectified:
The `verbose` argument for Python's `collections.namedtuple()` changed to keyword-only in >= Python3.6 (see https://docs.python.org/3/library/collections.html#collections.namedtuple)

This PR changes `debug.py` to reflect that.


kmike:
Thanks @objectified!

PR #80, opened 2018-03-29
added support for merging layer masks when using as_PIL()

kspes:
Hi,

I need to export layers from a PSD file and some of them have to have a layer mask (I can't change this requirement on my project). So I adjusted psd_tools code to multiply the alpha channel with the mask channel and then merge it all together when returning a PIL object.
it works good for me so far, so I'd like to share the code.

The code ain't great, some things could probably be done better, plus I may have not covered all situations due to my unfamiliarity with psd-tools codebase. However, I think it's a good start from which we can improve these commits and merge into psd-tools if you approve this approach.
The functionality could maybe be controlled by a global flag if someone doesn't want to merge layer masks?


kspes:
note, some of the continuous integration checks failed, but it seems only due to lack of python installations on the test machines.

kyamagu:
Resolved in 1.8.8

PR #172, opened 2019-12-26
Delete duplicate definition of has_mask function

meokz:
There is duplicate definition of has_mask function.

In Python, the earlier function definition is overwritten by the later function definition.

Unnecessary functions should be removed to avoid confusion.

I provide the modified source code.


kyamagu:
Yep, this is a bug

PR #183, opened 2020-02-14
Fix Overlay and Hard Light

bryab:
Overlay and Hard Light functions were producing incorrect results.  Test this by making a gradient layer, making another layer on top with various tones of gray, and compare psd_tools and photoshop's result.


kyamagu:
@bryab Thanks, but this does not pass tests

bryab:
Will have to take a look later.

kyamagu:
Blending implementation changed in v1.9.0.

Issue #279, opened 2021-08-12
Add support to release aarch64 wheels


odidev:
Problem
---------
On aarch64, ‘pip install psd-tools’ builds the wheels from source code and then installs it. It requires the user to have a development environment installed on his system. Also, it takes more time to build the wheels than downloading and extracting the wheels from PyPi.

Resolution
-----------
On aarch64, ‘pip install psd-tools’ should download the wheels from PyPi

@kyamagu, Please let me know your interest in releasing aarch64 wheels. I can help with this.

odidev:
PR raised for reviewing.

kyamagu:
Thanks!

PR #280, opened 2021-08-12
Add aarch64 wheel build support

odidev:
Added aarch64 wheel build support.
Related to https://github.com/psd-tools/psd-tools/issues/279, @kyamagu Could you please review this PR?

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -23,56 +23,64 @@ jobs:
       run: |
         python setup.py sdist
         twine upload --skip-existing dist/*
-  bdist_manylinux:
-    runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        container:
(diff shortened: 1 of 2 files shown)


kyamagu:
Hmm, qemu runs very slow on py36. Is there any workaround?

kyamagu:
While aarch64 support works in this suggested PR, maybe it is better to make the job extensible to other architecture. Also venv is not required in the CI environment.

Ref: https://github.com/actions/virtual-environments/issues/2095

Perhaps a more advanced tool like [`cibuildwheel`](https://cibuildwheel.readthedocs.io/en/stable/) would be a better option to go beyond the current minimum CI workflow.

kyamagu:
Ok, there is an error due to aggdraw not publishing cp39 for win64 wheel but otherwise looks fine.
I will merge this PR at this point, and open another PR to fix/refactor other CI tasks including the aggdraw issue.

PR #336, opened 2022-09-30
Fix clipping

### LINKED ISSUE #325: Group clipping into layer causes RecursionError
**Describe the bug**
Compositing results in a stack overflow when attempting to render set of layers containing a clipping group.

This stems from the fact that when opening, the group ("clipping" in the example file) is actually clipping into the layer that's inside itself ("red" in the example filed).

**To Reproduce**
```python
psd = PSDImage.open('test.psd')
psd[2] # Group('clipping' size=225x424)
psd[2][0].clip_layers # [PixelLayer('blue' size=225x424), Group('clipping' size=225x424)]
psd.composite(ignore_preview=True)
```

**Expected behavior**
The document should render without errors if there were no errors while opening.
Ideally, the file would also render as expected.

**File and screenshots**
This PSD file was created in Clip Studio Paint, which supports clipping groups into layers.
[example.zip](https://github.com/psd-tools/psd-tools/files/8957098/example.zip)

The group and layer order (the vertical red bar means the group/layer is clipping into the layer below):
layers

How Clip Studio Paint renders the file:
test

**Environment**
 - Python [3.10.4]
 - Version [v1.9.21]

**Additional notes**
I made some changes to the layer loading function to support this functionality, but I'm unsure if it would cause problems with other features or not. Here's the patch:
[recursive-clipping-layers.patch](https://github.com/psd-tools/psd-tools/files/8957688/recursive-clipping-layers.txt)

Compositing appears to produce the correct output as long as the document tree is properly parsed.

cromachina:
Fixes #325

### DIFF
--- a/src/psd_tools/api/psd_image.py
+++ b/src/psd_tools/api/psd_image.py
@@ -5,7 +5,7 @@ from __future__ import absolute_import, unicode_literals
 import logging
 
 from psd_tools.constants import (
-    BlendMode, Clipping, Compression, ColorMode, SectionDivider, Resource, Tag
+    BlendMode, CompatibilityMode, Compression, ColorMode, SectionDivider, Resource, Tag
 )
 from psd_tools.psd import PSD, FileHeader, ImageData, ImageResources
 from psd_tools.api.layers import (
@@ -39,6 +39,7 @@ class PSDImage(GroupMixin):
         self._record = data
         self._layers = []
         self._tagged_blocks = None
+        self._compatibility_mode = CompatibilityMode.DEFAULT
         self._init()
 
     @classmethod
@@ -439,6 +440,20 @@ class PSDImage(GroupMixin):
         """
         return self._record.layer_and_mask_information.tagged_blocks
 
+    @property
+    def compatibility_mode(self):
+        """
+        Set the compositing and layer organization compatibility mode. Writable.
+
+        :return: :py:class:`~psd_tools.constants.CompatibilityMode`
+        """
+        return self._compatibility_mode
+
+    @compatibility_mode.setter
+    def compatibility_mode(self, value):
+        self._compatibility_mode = value
+        self._compute_clipping_layers()
+
     def has_thumbnail(self):
         """True if the PSDImage has a thumbnail resource."""
         return (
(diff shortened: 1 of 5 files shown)


cromachina:
I have made it so that clipping on to a pass through layer causes no clipping to occur, which is the same behavior as SAIv2 and CSP. However it seems like this is not correct according to Photoshop, because it's causing a test to fail where a layer is clipped onto a pass through group.

cromachina:
I came up with a better way to deal with clipping layers than storing a clipping layer list in each layer when working on my own compositor, by simply storing the previous non-clipping layer's alpha, however, I'm not sure if I can easily translate it to `psd-tools`  compositor:

https://github.com/cromachina/psd-export/blob/9dd91aa39384bd201b2b4b7ada98dd4d0248be35/composite.py#L78-L106

kyamagu:
@cromachina What would be the test to check the correctness?

cromachina:
@kyamagu I don't have Photoshop so it's hard for me to confirm, but the test `test_layers.test_clip_adjustment` seems to be the offending one. From what I can infer from this test, it seems like in PS you can clip on to a `PASS_THROUGH` layer, but in SAI, the clipping behavior is ignored. The former would look like "source atop" and the latter would look like "source over". Does this sound correct?

src/psd_tools/api/psd_image.py
```diff
@@ -39,6 +39,7 @@ def __init__(self, data):
         self._record = data
         self._layers = []
         self._tagged_blocks = None
+        self._compatibility_mode = CompatibilityMode.DEFAULT
```
kyamagu, line 42: Actually, there is `VersionInfo` field in the resource section of a PSD file, that states the file creator like Adobe Photoshop. Perhaps this part should attempt at guessing the default compatibility mode.

```
from psd_tools.constants import Resource
version_info = psd.image_resources.get_data(Resource.VERSION_INFO)
```

src/psd_tools/constants.py
```diff
@@ -1,8 +1,20 @@
 """
 Various constants for psd_tools
 """
-from enum import Enum, IntEnum
+from enum import Enum, IntEnum, auto
 
+class CompatibilityMode(Enum):
```
kyamagu, line 6: Presumably there should be fixture files to test those. Any possibility?

src/psd_tools/api/psd_image.py
```diff
@@ -525,6 +540,38 @@ def _get_pattern(self, pattern_id):
                         return pattern
         return None
 
+    def _clear_clipping_layers(self):
+        for layer in self.descendants():
+            layer._clip_layers = []
+            layer._has_clip_target = True
+
+    def _compute_clipping_layers(self):
```
kyamagu, line 548: Appreciate if there is a short comment describing the behavior difference between tools

src/psd_tools/api/psd_image.py
cromachina: I can try something like that in bit.

PR #346, opened 2022-12-15
Accept other classId in Stroke constructor

Etienne-Gautier:
# Context
I have see a user-uploaded file that psd-tools can't read. This file was produced by Affinity Photo 1.10.4 and exported as PSD.

The file was failing in the `Stroke` constructor at `assert self._data.classID == b"strokeStyle"`. Instead the `classId` is `Strk`

Photoshop doesn't have a problem opening this file, and does so without warning. So the classId check can be either relaxed (accept 2 values) or removed entirely.

Just wondering if it is best to allow 2 values here, or to remove the assert altogether?

### DIFF
--- a/src/psd_tools/api/shape.py
+++ b/src/psd_tools/api/shape.py
@@ -164,7 +164,7 @@ class Stroke(object):
 
     def __init__(self, data):
         self._data = data
-        assert self._data.classID == b'strokeStyle'
+        assert self._data.classID == b'strokeStyle' or self._data.classID == b'Strk'
 
     @property
     def enabled(self):
--- a/src/psd_tools/api/shape.py
+++ b/src/psd_tools/api/shape.py
@@ -11,6 +11,7 @@ from __future__ import absolute_import
 import logging
 
 from psd_tools.psd.vector import Subpath, InitialFillRule, ClipboardRecord
+from psd_tools.terminology import Event
 
 logger = logging.getLogger(__name__)
 
@@ -164,7 +165,8 @@ class Stroke(object):
 
     def __init__(self, data):
         self._data = data
-        assert self._data.classID == b'strokeStyle' or self._data.classID == b'Strk'
+        if self._data.classID not in (b'strokeStyle', Event.Stroke):
+            logger.warning("Unknown class ID found: {}".format(self._data.classID))
 
     @property
     def enabled(self):


kyamagu:
The ID seems to be the [`Event`](https://github.com/psd-tools/psd-tools/blob/ac1ac5a9e70923037b550c76e3a6bdbba450d5ac/src/psd_tools/terminology.py#L1055) enum value. I do not know what values are possible here, but skipping the check does not sound safe. At least there should be a warning for unknown values.

Can you also come up with a test case?

Etienne-Gautier:
> I do not know what values are possible here, but skipping the check does not sound safe

I suggested removing the check because it seems like this is the only place where we assert for a specific classID in a constructor.

I'll try to generate a new file with this same issue, but the one I have right now has customer data and I can't share it

Etienne-Gautier:
> I'll try to generate a new file with this same issue, but the one I have right now has customer data and I can't share it

Still no luck generating a file with that issue despite trying a few different shapes and versions of this affinity software. Such files definitely exists in the wild.

Worst case, I can still get away with a monkey patch if psd-tools you won't accept this PR without a demo file (which I totally understand).

src/psd_tools/api/shape.py
```diff
@@ -164,7 +164,7 @@ class Stroke(object):
 
     def __init__(self, data):
         self._data = data
-        assert self._data.classID == b'strokeStyle'
+        assert self._data.classID == b'strokeStyle' or self._data.classID == b'Strk'
```
kyamagu, line 167: @Etienne-Gautier Can you do something like the following?

```python
from psd_tools.terminology import Event

if self._data.classID not in (b'strokeStyle', Event.Stroke):
    logger.warning("Unknown class ID found: {}".format(self._data.classID))
```

kyamagu: OK not to have a fixture file for now

Etienne-Gautier: Sure I'll make that change, I've also just managed to get down to a simple version of the file that reproduces the issue (zipped because github doesn't allow PSDs):
[user file cycled.psd.zip](https://github.com/psd-tools/psd-tools/files/10263567/user.file.cycled.psd.zip)

kyamagu:
The CI error is likely due to the outdated cibuildwheel version.

PR #350, opened 2023-01-17
support rgb float colors

Etienne-Gautier:
# Context
Some files are failing with  `KeyError: b'Rd  '` when parsing colors, this is due to the RGB components being encoded differently as floats (from 0 to 1) and under a different key in the descriptor. I have attached an example (in zip below).
[weird-rgb-encoding-2.PSD.zip](https://github.com/psd-tools/psd-tools/files/10430202/weird-rgb-encoding-2.PSD.zip).

This PR adds support for RGB colors encoded as floats


kyamagu:
LGTM, thanks

PR #359, opened 2023-06-17
Fix hue, sat, and vivid light

### LINKED ISSUE #233: Vivid Light not rendered correctly
Please see attached PSD.  Vivid light mode is not rendered correctly in psd_tools.  I am using the "new" composite, but I think the old one also has a (different) problem.

[psdtools_bug_vividlight.zip](https://github.com/psd-tools/psd-tools/files/4783044/psdtools_bug_vividlight.zip)

cromachina:
Fixes #233 

I found a minor order of operations bug in _set_sat and fixed that.

I also changed vivid light to use color burn and color dodge directly. This produces the exact same result as `[TS] Vivid Light` in Paint Tool SAI and `Vivid Light` in Clip Studio Paint, which is supposed to be the same as Photoshop, however I don't have Photoshop to verify. This change seems to contradict the original comment about using Photoshop's Hard Mix, but I think this produces the 'correct' result, or at least the result seems more correct (visually and aesthetically).

### DIFF
--- a/src/psd_tools/composite/blend.py
+++ b/src/psd_tools/composite/blend.py
@@ -293,7 +293,7 @@ def _set_sat(C, s):
     index_min = (C == C_min)
 
     index = index_mid & index_diff
-    B[index] = C_mid[index] - C_min[index] * \
+    B[index] = (C_mid[index] - C_min[index]) * \
         s[index] / (C_max[index] - C_min[index] + 1e-9)
     index = index_max & index_diff
     B[index] = s[index]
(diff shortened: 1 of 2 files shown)


cromachina:
I would update the test PSDs, but last time I tried exporting with SAI, the test suite did not like the resulting PSD file. It probably wouldn't be a good idea anyway if you are using Photoshop directly as a reference.

kyamagu:
@cromachina Thanks for the PR. It's difficult to let this PR merge without verifying the behavior in Photoshop.

cromachina:
For the hue and sat functionality, I did check against the adobe documentation, which is where I found the discrepancy for that function

image

kyamagu:
@cromachina Alright. Can you set the threshold value such that there is no testing error, or mark the test expected to fail?

PR #364, opened 2023-06-26
fix missing arguments when calling draw_solid_color_fill

Etienne-Gautier:
In my last PR [here](https://github.com/psd-tools/psd-tools/pull/361).

I missed a couple of call sites when adding an argument to `draw_solid_color_fill`.

This causes the following error:
```
TypeError: draw_solid_color_fill() missing 1 required positional argument: 'desc'
```

Sorry if anyone was affected by this bug in the 1.9.29 release (4 days old)

cc: @kyamagu


kyamagu (approved):
LGTM

PR #367, opened 2023-07-04
Add alternate 8ELE signiture for 8BIM tagged block

ekohilas:
Updates the tagged blocks metadata to accept the `8ELE` signature alongside `8BIM`.

I have also added a minimal sample test file with this issue, that only passes the tests with this change.

Thanks!


kyamagu:
Thanks!

PR #383, opened 2024-01-04
Migrate build tools to pyproject.toml

kyamagu:
Changes:
- Migrate build tools to `pyproject.toml`
- Use `pysen` for linting and formatting (no mypy support at this moment)
- Update CI configuration to the latest tools
- Deprecate tox


.github/workflows/ci.yml
```diff
@@ -52,28 +37,34 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        build: [cp37-*, cp38-*, cp39-*, cp310-*, cp311-*]
+        build: [cp38, cp39, cp310, cp311, cp312]
```
ksallee, line 40: any reason to remove cp37?

kyamagu: Python 3.7 is already EOL
https://devguide.python.org/versions/

Jerakin: I don't think that the version dev guide shouldn't be followed by libraries. A lot of industries moves a lot slower than being able to update python version every couple of years. Personally I think it only makes sense to remove the support for a python version if the library added something to make it not work in the old version.

ksallee: It's the problem I'm facing. Some clients are stuck on python 3.7 because they load Photoshop through an older version of ShotGrid Desktop, and doing the transition for all artists to the newer version of SG Desktop is not that easy. They're gonna bite the bullet to get this update, but TBH, it'd be better to retire these versions a bit later if possible? Media and entertainment industry is notoriously slow to update python versions (they were forced to switch to python 3 very recently if you can believe it!)

Jerakin: This happened to us too. We rely on 3.7 and had to lock psd-tools down to `1.2.28`

kyamagu: The older releases are always available. Just pin the version to `psd-tools<=1.9.28` for whatever reason. The dependent packages (e.g., `Pillow`, `scipy`) already do not support deprecated Python versions in the latest release.

There will be `sdist` for the unsupported environment. You are free to build and host extra wheel packages.

PR #386, opened 2024-01-05
fix: include pyx file in build

Jerakin:
v1.9.29 is broken as the pyx file is not included in the build

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,1 @@
+include src/psd_tools/compression/_rle.pyx
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,7 +50,7 @@ psd-tools = "psd_tools.__main__:main"
 where = ["src"]
 
 [tool.setuptools.package-data]
-mypkg = ["*.pyx"]
+psd_tools = ["*.pyx"]
 
 [tool.setuptools.dynamic]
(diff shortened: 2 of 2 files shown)


kyamagu:
Right, thanks!

Jerakin:
Might be. Which platform doesn't it work on? I have only tried this on Windows (only platform I got access to locally).

kyamagu:
See [the build log for sdist](https://github.com/psd-tools/psd-tools/actions/runs/7420048663/job/20196240242?pr=386), [which is Ubuntu](https://github.com/psd-tools/psd-tools/blob/main/.github/workflows/ci.yml).
The build log indicates `*.pyx` file is not copied into a package for some reason.

I can confirm that this PR also works as expected in my local macOS. This is likely a CI issue.

Jerakin:
Maybe there's a bug a manifest file is needed. Shouldn't be needed anymore but worth a try.

kyamagu (approved):
LGTM. This might be a bug or a version issue of build tools.

Jerakin:
> BTW, is there any runtime issue? The wheel packages contain a compiled binary.

Not run time, on windows you can't even pip install it. When it is getting the requirements to build wheel it will throw an `ValueError: 'src/psd_tools/compression/_rle.pyx' doesn't match any files`

PR #392, opened 2024-02-21
reworked packbits/rle algorithms

oh-ok:
Hi! Thanks for developing this awesome tool. 

I've been looking for ways to compress PSDs while still having them natively readable by Photoshop, and I came up with a slightly more space efficient PackBits encoding algorithm. It admittedly only saves a tiny amount of space compared to Photoshop's implementation, but I thought it was still worth making a PR here.

I also took the time to refactor the rest of the RLE module here, since it was possible to simplify and optimise the code a bit. Most of the performance gains are from removing the overhead of calling extra Python functions.

You'll also notice that I swapped out raw C strings for C++ strings in the Cython code, this just makes it far easier to sit back and let Cython and C++ manage the memory and type conversions. If you'd prefer to use raw C strings, I could refactor this to do that instead.

In terms of speed, I ran some tests with `timeit`:
```python
timeit.timeit(
    'm.encode(t)',
    setup='import {module} as m; t=bytes.fromhex("aaaaaa80002aaaaaaaaa80002a22aaaaaaaaaaaaaaaaaaaa")',
    number=10_000_000
)
timeit.timeit(
    'm.decode(t, 24)',
    setup='import {module} as m; t=bytes.fromhex("feaa0280002afdaa0380002a22f7aa")',
    number=10_000_000
)
```

(Running on an M1 MacBook Pro)

| Module | Encode Time (s) | Decode Time (s) |
|--------|--------|--------|
| `.py` old | 50.8301086375 | 22.117071729000052 |
| `.py` new | 42.34482602520001 |16.344653579000003 |
| `.pyx` old | 15.525218291600003 | 4.582728499999973 |
| `.pyx` new | 1.6874734792 |  1.6621730334000002 | 

In terms of space saved, tested on an 1024x1024 image with 1 layer consisting of a unique rendered cloud on each channel, and some Lorem Ipsum text over the top.
For psd_tools RLE and new RLE, the result is the length returned by `image_data.set_data`
Ps RLE is `len(psd.image_data.data)` where `psd.image_data.compression == Compression.RLE` 

|Ps RLE|psd_tools RLE|new RLE| 
|--------|--------|--------|
|2836621|2951643|2816249|

Which translates to about a 4.58% decrease in file size compared to psd-tools' previous algorithm, and a 0.72% decrease compared to Ps' implementation

Let me know if you have any feedback.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ __pycache__/
 # C extensions
 *.so
 *.c
+*.cpp
 
 # Distribution / packaging
 .Python
--- a/src/psd_tools/compression/_rle.pyx
+++ b/src/psd_tools/compression/_rle.pyx
@@ -4,12 +4,6 @@
 from libcpp.string cimport string
 
 
-class PackBitsDecodeError(Exception):
-    def __init__(self, result, *args: object) -> None:
-        self.result = result
-        super().__init__(*args)
-
-
 def decode(const unsigned char[:] data, Py_ssize_t size) -> string:
     """decode(data, size) -> bytes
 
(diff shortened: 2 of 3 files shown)


kyamagu:
Thanks!

PR #412, opened 2024-06-13
Add a new property to SmartObject: transform_box

dario-spagnolo:
I suggest this PR to expose and document a new property for SmartObject: `transform_box`

It comes from the `transform` property of the PlacedLayerData TaggedBlock. It is a tuple with 8 items (x1, y1, x2, y2, x3, y3, x4, y4) where 1 is top left corner, 2 is top right, 3 is bottom right and 4 is bottom left.

I chose to rename it in order to avoid confusion with the `transform` property of TypeLayer which is a tuple with 6 items (xx, xy, yx, yy, tx, ty) and is an affine transformation matrix.

### DIFF
--- a/src/psd_tools/api/smart_object.py
+++ b/src/psd_tools/api/smart_object.py
@@ -28,6 +28,12 @@ class SmartObject(object):
                 self._config = layer.tagged_blocks.get_data(key)
                 break
 
+        self._placed_layer = None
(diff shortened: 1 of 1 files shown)


kyamagu:
Ok, LGTM

kyamagu (approved):

PR #418, opened 2024-06-27
New PSDImage method to set the logging level

dario-spagnolo:
Prior to these changes, the logging level could only be set when running psd-tools from the command line.

Now, it is possible to set the logging level when using psd-tools as a module.

### DIFF
--- a/src/psd_tools/api/psd_image.py
+++ b/src/psd_tools/api/psd_image.py
@@ -602,7 +602,7 @@ class PSDImage(GroupMixin):
         Default logging level is WARNING.
         """
         if log_level in ("CRITICAL", "ERROR", "WARNING", "INFO", "DEBUG"):
-            logging.basicConfig(level=log_level)
+            logger.parent.setLevel(log_level)
         else:
             logging.warning(f"Logging level '{log_level}' is not valid.")
 
--- a/src/psd_tools/api/psd_image.py
+++ b/src/psd_tools/api/psd_image.py
@@ -593,6 +593,19 @@ class PSDImage(GroupMixin):
 
         rec_helper(self)
 
+    def set_log_level(self, log_level):
+        """
+        Set the logging level.
+
+        :param log_level: string of the desired logging level 'CRITICAL', 'ERROR', 'WARNING', 'INFO' or 'DEBUG'.
+
+        Default logging level is WARNING.
+        """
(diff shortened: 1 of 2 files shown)


src/psd_tools/api/psd_image.py
```diff
@@ -593,6 +593,19 @@ def rec_helper(layer):
...
+    def set_log_level(self, log_level):
+        """
+        Set the logging level.
+
+        :param log_level: string of the desired logging level 'CRITICAL', 'ERROR', 'WARNING', 'INFO' or 'DEBUG'.
+
+        Default logging level is WARNING.
+        """
+        if log_level in ("CRITICAL", "ERROR", "WARNING", "INFO", "DEBUG"):
+            logging.basicConfig(level=log_level)
```
kyamagu, line 605: There is actually a caveat in resetting the logging configuration once set:
https://stackoverflow.com/questions/12158048/changing-loggings-basicconfig-which-is-already-set

Maybe adding `force=True` option is fine, but that would destroy all the other handlers set outside of the psd-tools. https://docs.python.org/3.8/library/logging.html?highlight=logging#logging.basicConfig

kyamagu (changes requested):

src/psd_tools/api/psd_image.py
dario-spagnolo: You are right. 
In the latest commit I added a logger at the "psd_tools.api" level which serves as the local root of psd_tools when used as a module. 
The new set_log_level() method only impacts the log level of "psd_tools.api" and its children, which is effectively every logger declared inside src/psd_tools/api but nothing outside of it.

dario-spagnolo: The commit message should read "Better way to set the log level without impacting **loggers** outside of psd-tools" (not handlers)

dario-spagnolo:
You are definitely right and these 2 lines should be enough to set the log level from the application using psd-tools. 

The reason I added the set_log_level() method in the first place is because I didn't know I could set the log level from outside psd-tools. I was expecting some way to configure the psd-tools module and make it more verbose, like it can be done on the CLI version.

Maybe the set_log_level() can stay and be helpful to other users who are not used to the logging module. But if you think it's redundant and not the right way to go, just discard this PR, I'm ok with that :)

kyamagu:
@dario-spagnolo Alright, I'd close this PR since this seems the application matter. Thanks anyway for the suggestion

PR #419, opened 2024-06-27
Add new property text_type for TypeLayer

dario-spagnolo:
Type layers can contain texts of 2 types : point (character) or paragraph (area).
The information is stored in EngineDict and this PR introduces a new property `text_type` for TypeLayer. 
It also introduces a new TextType constant.


kyamagu (approved):
LGTM

Issue #84, opened 2018-07-30
Writing a new layer


gtm2122:
Is there any way to write into a psd object a new layer ? for example if I just want to append a white mask onto the existing list of psd_tools.Layer s and then save it as a new psd file ? I could not find an encoding method

kyamagu:
Currently writing support is not available.

genemiller:
Dear kyamagu,
I appreciate that you are working on Writing/Editing support. Meanwhile, do you know of other programmatic or scripting solutions that are available to edit layer content right now? I am trying to replace images and text for selected layers in existing PSD files. I started looking at (a) Adobe's Photoshop Scripting in javascript, and (b) ImageMagick command line tool. I am experienced in python and writing shell scripts, but I never used javascript. Do you have any opinions?
Thanks again for a very nice tool.

kyamagu:
@genemiller Adobe's official scripting API is the only reliable option.
https://www.adobe.com/jp/devnet/scripting.html

Also there seems what's called Adobe CEP right now, but I have never played with that.
https://github.com/Adobe-CEP

In any case, Adobe's tools are proprietary and seems requiring specific scripting knowledge.

genemiller:
Thank you. I will check back at a later time to see how Writing/Editing support is progressing. Meanwhile I can follow javascript examples for Adobe's Photoshop Scripting. (Adobe CEP is is used to create plugins, which is overkill. And ImageMagic seems useless for what I need.)

rivergold:
Is there any way to change the raw image data in the layer of psd ? I want to do some image processing，such as resize, move and rotate, and save it into a new psd. Is it possible?

kyamagu:
@rivergold Nope, writing support is not available

hazar-e:
hi @rivergold  
> Is there any way to change the raw image data in the layer of psd ? I want to do some image processing，such as resize, move and rotate, and save it into a new psd. Is it possible?

here is simple script, saves the numpy image list as psd layers. (only rgb, rgba images, not masks etc)
https://gist.github.com/hazar-e/6a7ebcd3c267b0245456cc9d9cf6bcdf

**pip3 requirements;**

- pytoshop
- numpy
- psd_tools

**usage;**

```
import cv2
from ps import savePSD, openPSD


inputpsd = openPSD("input.psd")

# resize bottom layer of psd 
[...]

Issue #237, opened 2020-07-01
Low quality and color shift exporting CMYK psd to jpg/png


bvb-sc:
Hi,

When executing the following command, a low quality and color shifted result is rendered: 

`psd-tools.exe export "cmyk.psd" "cmyk.jpg"` 

**Result**:

![image](https://user-images.githubusercontent.com/67702686/86270389-742bbc80-bbcb-11ea-8504-273d9a01b431.png)

**Expected result**:

![image](https://user-images.githubusercontent.com/67702686/86270419-80b01500-bbcb-11ea-8855-74de936d6723.png)

**Reference file**:

[cmyk.zip](https://github.com/psd-tools/psd-tools/files/4859213/cmyk.zip)

**Version info**:
OS: Windows
Python: 3.8.3
psd-tools: 1.9.13

**Additional info**:

Originally we wanted to export directly to png, but this resulted in the error below. The error makes sense, since png does not support CMYK and psd-tools doesn't seem to support native conversion from CMYK to PSD. Looking into PIL's Image & ImageCms features through scripting, we found color management methods, but they also result in a low quality and color shifted rendition.

```Traceback (most recent call last):
  File "XXX\python\python38\lib\site-packages\PIL\PngImagePlugin.py", line 1179, in _save
    rawmode, mode = _OUTMODES[mode]
KeyError: 'CMYK'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "XXX\python\python38\lib\runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "XXX\python\python38\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "XXX\Python\Python38\Scripts\psd-tools.exe\__main__.py", line 7, in <module>
  File "XXX\python\python38\lib\site-packages\psd_tools\__main__.py", line 58, in main
    image.save(args['<output_file>'])
  File "XXX\python\python38\lib\site-packages\PIL\Image.py", line 2134, in save
    save_handler(self, fp, filename)
  File "XXX\python\python38\lib\site-packages\PIL\PngImagePlugin.py", line 1181, in _save
    raise OSError("cannot write mode %s as PNG" % mode)
OSError: cannot write mode CMYK as PNG```

Any help on how to correctly convert psd to png, without color shifts/quality loss would be appreciated.

kyamagu:
[...]

PR #428, opened 2024-08-24
Layer structure edition and Group/PixelLayer Creation

al3ks1s:
- Added _update_record() function that will read the high level layers structure and recompile the LayerRecords and ChannelImageData from it. This function is called in every PSDImage function that uses the records to produce data 72dd34452802ea94c520735d96e943c052368e57
  - Added the _bounding_record and _bounding_channels for the Group layers, they hold the records with the SectionDivider.BOUNDING_SECTION_DIVIDER so that we don't have to recompute them. (Renamed from _open_record/_channels in baef8cafce452b428e276541ac94eca3a3275071)
  - Added methods for layer structure modifications 631d23f2bd4dc6de7dedfcde24fc7a69b85be98f : 
    - Addition to a group (Group.add_layer)
    - Deletion 
    - Moving to a group (different context from "addition to group", Layer.move_to_group)
    - Moving up and down within the group
- Group creation fef6ca308a66a9b0b895b449b3e83c6d9d981902 : 
  - Minimal standalone Group layer using "new", add it to the structure using the methods above.
  - Create a group from a list of layers. The group will insert itself in place of the layer at index 0 or under the "parent" layer given in parameter.
- PixelImage creation 1d8f6c758aeeb2439894721b6900f07699528ae3 :
  - PixelImage layer creation from a PIL object, for now limited to RGBA mode.

Edit : Known bug: when adding a layer, whether group or pixel layer with 4 channels, a psd with 3 channels in header might not be possible to open in Photoshop

https://github.com/psd-tools/psd-tools/issues/84


kyamagu:
Hi, can you also create test cases in this PR? This is going to be a significant change that needs thorough tests and error checks.

al3ks1s:
Hello, i added tests for the functions for the edition methods, tell me if those are lacking anything. Also the PixelLayer.frompil test isn't doing anything because i'm going to rework the function for handling all images modes

kyamagu:
There are some other things to consider when editing the structure.

- State to invalidate the composite cache: Unless `force=True,` by default, the composite function uses the cached rendered view (available when Photoshop saves PSD in the compatibility mode). Any editing will invalidate the cache, and the composite function needs a state flag to control this behavior.
- PSB format compatibility: PSB files keep layer information in the other section. Needs compatibility testing with Adobe Photoshop.
- (Minor) Image Resources section seems to include some information on layer IDs, which is likely used in Adobe Photoshop. Needs to inspect how they change when layers are added/deleted/reordered.

al3ks1s:
When creating a layer with a bitmap pil mode, the bit depth of 1 for a psd of depth 8 create an issue when fetching the data of a channel through get_data, the data length being 8 times smaller than expected. Should the pil image be converted to grayscale through creation or add a depth attribute to the layer class, keeping the original data?

src/psd_tools/api/layers.py
```diff
@@ -812,7 +1034,57 @@ class PixelLayer(Layer):
...
+
+        layer_record = LayerRecord(top=top, left=left, bottom=top + pil_im.height, right=left + pil_im.width)
+        channel_data_list = ChannelDataList()
+
+        layer_record.name = layer_name
+        layer_record.channel_info = [ChannelInfo(ChannelID.TRANSPARENCY_MASK, 2)]
+        
+        channel_data_list.append(ChannelData(compression))
+
+        print(pil_im.mode)
```
kyamagu, line 1063: Use `logger`

kyamagu:
ChannelData expects the format specified in [the file header](https://www.adobe.com/devnet-apps/photoshop/fileformatashtml/#50577409_pgfId-1055726). If the PSD has a Bitmap color mode, the channel data should have the expected binary data in big-endian. If PIL has a different format, it must be converted to whatever format PSD expects.

src/psd_tools/api/layers.py
al3ks1s: Whoops, forgot to clean that up

al3ks1s:
> Hello,
> 
>     * i've added a state attribute to the psd class (_updated_layers) which is used by _update_record to disable updating it if no layer changed was made and force composite function to ignore cache (and added the necessary functions in the layer classes to set the flag to true when using a layer edition function, it might be an idea to make this a decorator). Not sure it's the best solution available
> 
>     * I looked into the PSB format and i can't create a PSB compatible with my installation of photoshop, _but_ it happens whether i update the layers (and record) or not so i can't properly test it. Photopea can open it though.
> 
>     * I'll look into that

On the PSB part,  the issue wasn't the psb itself but the ZIP compression on the imagedata, though psd_tools fails to produce an image big enough with RLE on imagedata

al3ks1s:
> ChannelData expects the format specified in [the file header](https://www.adobe.com/devnet-apps/photoshop/fileformatashtml/#50577409_pgfId-1055726). If the PSD has a Bitmap color mode, the channel data should have the expected binary data in big-endian. If PIL has a different format, it must be converted to whatever format PSD expects.

Refactored some of the code for layer structure edition, and added something to properly distribute the _psd attribute across all child layers when moving them from one place to another. (Useful when moving groups with a lot of levels to keep data consistent)

The PSB works with all this, as long as psd_tools is able to create one to begin with (The process died when trying to generate one with RLE compression and ZIP made it so my photoshop version couldn't open it)

I only quickly looked at the Image resource part and it seems the slices resources are the only relevant resources that needs the layer id. I will investigate later.

As for the image mode thing, i added the possibility to use a psd_file as a reference to convert the pil image when creating the layer from it, it might be needed to make this parameter mandatory as a pixel layer with more channels than the psd document has cannot be opened (Photoshop outputs an early EOF error). With an automatic conversion when moving a pixel layer to the document or something.

al3ks1s:
Slices are used mainly (if not solely) for web design, to cut some images and insert them into web pages, i heavily doubt its _that_ used, and hence, that urgent to take care of it

al3ks1s:
I was thinking of adding a color mode attribute to the layer class, so that they can be independently composited, and then whenever moving the layer to a new psd, the color mode would change to the psd's and a pixel layer would be converted to the psd's color mode, what are your thoughts on this?

kyamagu:
@al3ks1s 
> adding a color mode attribute to the layer class, so that they can be independently composited

I don't understand what this means. Are you trying to composite the grayscale, RGB, or CMYK layers if they coexist in the same PSD?

al3ks1s:
The idea was that when creating a layer without being linked to a psd, the layer cannot be converted to PIL image, but adding a color attribute isn't actually necessary so nevermind.

al3ks1s:
I made a conversion function but still face a few issues with generating a working psd when making a psd with modes that have no explicit alphas, a psd in RGB mode will allocate 3 channels (instead of like 4 for RGBA). i'll add a property in the psd_image to set whether the psd_image was made with a mode containing an alpha channel or not

al3ks1s:
It was actually easier to implement than i thought first. as i said, i added a pil_mode to the psd_image that will recompute the original image mode the object was created with, taking the number of channels and the color_mode, and changed the frompil function so that the image would be converted to the format needed by the psd file to make sure there are no early eof issues or some other non-descriptive errors.

I don't think the conversion function needs a test case because it makes use of the frompil function from pixellayer. (But if you think otherwise i can add one)

al3ks1s:
I still face issues when making images on psds that have no alpha channel, i don't understand whats the issues on photoshop's side that makes it unable to open the file. And photoshop lack of explicit error message doesn't help

al3ks1s:
All in all, it works but i lost a lot of time on a side effect of the conversion

al3ks1s:
CMYK Colors are botched in the layer, i'll investigate later

al3ks1s:
Using information described here https://github.com/psd-tools/psd-tools/issues/237, cmyk images needs to be inverted when converting to layer data to get the proper output, in both the compsite and topil methods and the normal opening of a psd in photoshop

al3ks1s:
Met blocking edge case on the RLE encoding, I get the following error : IndexError: Out of bounds on buffer access (axis 0) on line 90 

- elif ((j+2 == length) or (MAX_LEN - (j - i) <= 2)) and (data[j] == data[j+1]):

kyamagu:
@al3ks1s Great catch! Are you sure the fix can reconstruct the same raw data in the edge case?

al3ks1s:
Took a look at the images before and after the conversion with the fix in photoshop with difference mode, seen nothing so i'd say it's good.

kyamagu:
It's best to have a test fixture for this corner case:

```python
data = b'xxx'
assert data == decode(encode(data), len(data))
```

al3ks1s:
I checked with several images that previously failed and saw no issue with them, but yeah i'll make a test case with this specific type of data.

al3ks1s:
I extracted a sample of the data that was failing and added it to the rle identity test.

al3ks1s:
Any ideas of things still left to consider or feedback on the code or tests? The one thing i'm working on right now is to block the possibility of creating reference loops in the layer structure

src/psd_tools/api/layers.py
```diff
@@ -561,6 +627,52 @@ def bbox(self):
...
+        :param layer: The layer to add
+        """
+
+        assert layer is not self
+
+        self.add_layers([layer])
+
+        return layer
+
+    def add_layers(self, layers = []):
```
kyamagu, line 643: Default argument should not be a mutable object. In this case drop the default arg

src/psd_tools/api/layers.py
```diff
@@ -561,6 +627,52 @@ def bbox(self):
...
+            layer._parent = self
+            self._layers.append(layer)
+
+        _psd = self if self.kind == "psdimage" else self._psd
+
+        for layer in self.descendants():
+            
+            if layer._psd != _psd and _psd is not None:
+                if layer.kind == "pixel":
+                        layer._convert(_psd)
```
kyamagu, line 667: Incorrect indent. Apply a code formatter

src/psd_tools/api/layers.py
```diff
@@ -576,6 +688,22 @@ def __setitem__(self, key, value):
     def __delitem__(self, key):
         return self._layers.__delitem__(key)
 
+    # Use the layer structure methods instead, they make more comprehensive changes (with metadata updates)
```
kyamagu, line 691: I'm not sure what these methods are intended for. Are you trying to emulate container types?
https://docs.python.org/3/reference/datamodel.html#emulating-container-types

src/psd_tools/api/layers.py
```diff
@@ -664,8 +792,15 @@ def _get_bbox(layer, **kwargs):
         lefts, tops, rights, bottoms = zip(*bboxes)
         return (min(lefts), min(tops), max(rights), max(bottoms))
 
-    def __init__(self, *args):
-        super(Group, self).__init__(*args)
+    def __init__(self, psd, record, channels, parent, _bounding_record = None, _bounding_channels = None):
```
kyamagu, line 795: `_bounding_record` and `_bounding_channels` do not look friendly to the API users. Although instantiating a `Group` object is not likely to be a common case, I would suggest splitting the setter method for those private data.

kyamagu:
On the design of layer APIs, one consideration might be to check the existing Photoshop scripting APIs:
https://objjob.phrogz.net/pshop/hierarchy

Although there is no need to implement the identical APIs, some of the operations could follow the design pattern.

kyamagu:
Also don't forget to mark the comments resolved when fixed

src/psd_tools/api/layers.py
```diff
@@ -812,7 +1013,79 @@ class PixelLayer(Layer):
...
+        :param compression: Compression algorithm to use for the data.
+        :param psdfile: The psd file the image will be converted for.
+
+        :return: A :py:class:`~psd_tools.api.layers.PixelLayer` object
+     
+        """
+
+        assert pil_im
+
+        if pil_im.mode == "1":
```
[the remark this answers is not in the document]
al3ks1s: Yes, as long as the psd is provided it will convert the image to the proper mode. By the way, writing a psd with layers in other color modes than the PSD expects doesn't seems to hinder photoshop when opening the psd file. Photoshop itself will make the conversion (though i think its cleaner to force it on the images in the code)

src/psd_tools/api/layers.py
```diff
@@ -812,7 +1013,79 @@ class PixelLayer(Layer):
...
+            pil_im = pil_im.convert("L")
+       
+        if psd_file is not None:
+            pil_im = pil_im.convert(psd_file.pil_mode)
+        else:
+            logger.warning("No psd file was provided, it will not be possible to convert it when moving to another pdf. Might create corrupted psds.")
```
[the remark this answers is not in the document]
al3ks1s: I should review what i write more.

src/psd_tools/api/layers.py
al3ks1s: I had left that because the layers variable itself isn't being written into when the default case occurs but true i should have removed it.

src/psd_tools/api/layers.py
al3ks1s: Unfortunate tab key press...

src/psd_tools/api/layers.py
```diff
@@ -812,7 +1013,79 @@ class PixelLayer(Layer):
...
+        
+        channel_data_list.append(ChannelData(compression))
+        channel_data_list[0].set_data(b'\xff'* (pil_im.width * pil_im.height), pil_im.width, pil_im.height, get_pil_depth(pil_im.mode.rstrip("A")))
+
+        for channel_index in range(get_pil_channels(pil_im.mode.rstrip("A"))):
+            
+            channel_data = ChannelData(compression)
+
+            # Need checking for bitmap, or bitmap conversion, since depth is global to PSDImage, converting to grayscale in such case
+            channel_data.set_data(pil_im.getchannel(channel_index).tobytes(), pil_im.width, pil_im.height, get_pil_depth(pil_im.mode.rstrip("A")))
```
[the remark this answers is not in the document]
al3ks1s: What do you mean by "multiple color modes"? This channel conversion works for all color modes supported by psd_tools tested here : https://github.com/al3ks1s/psd-tools/blob/086e35cde86001a7d0fd508065e583099ef4f565/tests/psd_tools/api/test_layers.py#L310

I will add tests for multichannel and indexed modes

src/psd_tools/api/layers.py
al3ks1s: That was the goal yes, i first added them so that groups would be assimilated to python lists, and i use some of them in other structure modification methods, but i'm now not sure they're that relevant to have.

al3ks1s: Do you think i should move the add_layer(s) methods (and similar ones) to the corresponding container emulation ones or should i just remove them ?

src/psd_tools/api/layers.py
al3ks1s: I added the new() method for creating a new group, as the __init__ needs several types of data to be already created (records and channels for example). new() creates all those necessary datas and the two attributes you mentionned.

The only other instance of direct group instantiation is in PSDImage._init for which these data already exist.

Though now that i think about it, the only place where this is absolutely needed is in the _update_record, and we can compute them in this function. I need to check if there's any important data in these record and channels of the folder.

al3ks1s: there are layer ids and other metadata that it's probably best to keep around for compatibility and data consistency

src/psd_tools/api/layers.py
kyamagu: Container emulation is optional. I would suggest either of the following:

1. Implement the public method name, e.g., `append`, instead of the private name `_append`, following [the standard data structure](https://docs.python.org/3/tutorial/datastructures.html#more-on-lists)
2. Drop those private methods

src/psd_tools/api/layers.py
kyamagu: Okay, if there are still cases for keeping those records, drop the arguments from the `__init__` and add a private setter method for them

src/psd_tools/api/layers.py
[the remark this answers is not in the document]
kyamagu: As far as I remember, CMYK must invert the channel data to keep PIL and photoshop compatible. Need to check if there are other such cases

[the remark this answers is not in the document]
al3ks1s: CMYK inversion is taken care of, i'll check indexed and multichannel now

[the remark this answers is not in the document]
al3ks1s: Seems like indexed and multichannel aren't supported at psd creation.

src/psd_tools/api/layers.py
al3ks1s: Removed them from the constructor and moved it to _set_bounding_records

src/psd_tools/api/layers.py
al3ks1s: I went with the container emulation suggestion for the GroupMixin class, the only two list methods i didn't make are sort and reverse which i don't think are very relevant for layer groups. I also added the relevant tests for these methods.

src/psd_tools/api/layers.py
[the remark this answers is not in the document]
kyamagu: Ok, then validate and raise an error if the color mode is not supported

kyamagu:
@al3ks1s Ok, it looks good overall. The CI error is from one of the dependencies in the macOS runner and should be fixed in the other PR. The last thing is to edit documentation. Can you make sure that all public methods get a docstring?

Edit to the user documentation (docs/) can be in another PR.

al3ks1s:
I can add it in a few hours and include it in this pr if needed

kyamagu:
Great. Could you edit the user documentation to accommodate this update? Thanks!

al3ks1s:
I added the usage documentation needed, tell me if it is lacking anything/isn't clear enough. Also tell me if anything is missing in the docstrings, i will fix tomorrow.

kyamagu (approved):

kyamagu:
@al3ks1s Many thanks!

al3ks1s:
I forgot to edit the readme to add the layer structure edition in the supported features

kyamagu:
Alright, I'm now prepping for the `1.10.0` release. I will edit the README in that PR.

Issue #433, opened 2024-09-25
PixelLayer conversion when the pixel layer is empty


al3ks1s:
Hi, 

When going through a script for fusing psds, i encountered a bug concerning the pixel layer conversion.

Basically, when calling self.topil() in _convert, an entirely empty pixel layer with no data will send back a `None` object. Causing the conversion to a new color mode to fail in PixelLayer.frompil.

Do you think the topil() in the conversion should be changed to composite() (as composite returns a proper pil object) or modify topil to send back an empty image?

kyamagu:
`topil` is intended for exporting the raw pixel data, and if there is no real channel data in the layer, it returns `None`. If you need to ensure the output is always an image object, yes, you should use `composite`.

al3ks1s:
I will just add a check in the conversion function that will handle this special case.
Along a few minor changes that I thought of afterward

PR #435, opened 2024-10-01
Conversion issue fix, Layer metadata refactor, effect pattern transfer

al3ks1s:
Hi, 

I made the fix for the conversion issue mentioned in https://github.com/psd-tools/psd-tools/issues/433, i went with the composite() function.

I refactored the functions for layer validity verification and metadata update.
In the latter i added something to retrieve the effect patterns from a psd to add to the new one when moving a layer around. As i encountered an issue for psd opening concerning this. There may be other things to consider in this case, I guess those will be found along the way.

I also added functions for searching layers by name. Yielding either the first match, or a generator of all of them.

### DIFF
--- a/src/psd_tools/api/layers.py
+++ b/src/psd_tools/api/layers.py
@@ -569,8 +569,6 @@ class Layer(object):
         self.parent.remove(self)
         self.parent.insert(newindex, self)
 
-        self.parent._update_psd_record()
-
         return self
 
     def move_down(self, offset = 1):
@@ -608,11 +606,12 @@ class Layer(object):
             
             psd_global_blocks.get(Tag.PATTERNS1).data.extend(
                 [
-                    pattern for pattern, pattern_id in zip(sourcePatterns, pattern_ids) if 
-                        pattern_id == pattern.pattern_id and 
-                        pattern_id not in [targetPattern.pattern_id for targetPattern in psd_global_blocks.get(Tag.PATTERNS1).data]
+                    pattern for pattern in sourcePatterns if 
+                        pattern.pattern_id in pattern_ids and 
+                        pattern.pattern_id not in [targetPattern.pattern_id for targetPattern in psd_global_blocks.get(Tag.PATTERNS1).data]
                 ]
             )
+
 
 class GroupMixin(object):
     @property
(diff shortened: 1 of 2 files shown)


src/psd_tools/api/layers.py
```diff
@@ -810,6 +820,28 @@ def descendants(self, include_clip=True):
...
+    def find(self, name):
+        """
+        Returns the first layer found for the given layer name
+        
+        :param name:
+        """
+
+        for layer in self.descendants():
+            if layer.name == name:
+                return layer
```
kyamagu, line 832: Why not reusing the iterator method?

    for layer in self.findall(name):
        return layer

src/psd_tools/api/layers.py
```diff
@@ -779,6 +757,38 @@ def count(self, layer):
...
+
+        _psd = self if self.kind == "psdimage" else self._psd
+
+        for layer in self.descendants():
+            
+            if layer._psd != _psd and _psd is not None:
+                if layer.kind == "pixel":
+                    layer._convert(_psd)
+            
+                layer._fetch_patterns(_psd)
```
kyamagu, line 785: While I understand the logic, I am not sure why this method only supports patterns. What is the plan for tagged block items? Is the approach here future-proof when considering other items?

al3ks1s: I only made the method for patterns because it's the only one encountered during a workflow, I have yet to find another global tagged block (because the patterns are stored in the layer_and_mask_info tagged blocks not in the layer's blocks) that would need to be copied from one psd to another and which conditions should it meet for it. The Adobe spec isn't very explicit on whether it uses a layer's tagged blocks or the global one to store data.

Now when it comes to future proofing, I can probably move part of the pattern fetching logic to the pattern or effect class other than that without more informations on what datas need to be moved I can't really tell

kyamagu: Ok, perhaps it would be better to have a method like `_fetch_tagged_blocks` and handle patterns inside of it for the future

al3ks1s: I renamed the function into `_fetch_tagged_blocks`, i will leave its body the same for now and whenever a new type of data needed to be moved i'll refactor it to be more general.

kyamagu:
@al3ks1s Thanks!

Issue #437, opened 2024-10-08
AttributeError when saving PSDImage after adding PixelLayers to PSD with Artboard


tuxedocat:
## Describe the bug

When appending/inserting PixelLayers to Artboards and attempting to save it, an `AttributeError` is raised: `'NoneType' object has no attribute '_lengths'`.
Adding PixelLayers to Groups is fine.

## To Reproduce

```python
from psd_tools import PSDImage
from psd_tools.api.layers import PixelLayer
import PIL.Image
import tempfile

# Open the artboard PSD file
base = PSDImage.open('tests/psd_files/artboard.psd')

# Create a new PIL image
pil_rgb = PIL.Image.new("RGB", (30, 30))

# Create a new pixel layer from the PIL image
pixel_layer = PixelLayer.frompil(pil_rgb, base)

# Append or insert the pixel layer to the artboard
base[0].append(pixel_layer)

# Attempt to save the modified PSD
with tempfile.TemporaryDirectory() as tmpdir:
    base.save(f"{tmpdir}/modified_artboard.psd")
```

## Expected behavior

The new pixel layer should be added to the artboard, and the modified PSD should be saved without any errors.
Adding pixel layers to the file without artboards, i.e. adding them to groups, is OK.

## Environment

- CPython: 3.12.3
- psd-tools version: 1.10.0 
    - also confirmed on current [HEAD](https://github.com/psd-tools/psd-tools/tree/4214fc9ef8d20de0579371bf1a99d268ef593e6f)

## Additional context

Found this when adding test cases to `test_layers.py`.

```python
@pytest.mark.parametrize(
    "filename",
    [
        "1layer.psd",
        "2layers.psd",
        "artboard.psd",
    ],
)
def test_add_pixel_layer_and_write(filename: str, pixel_layer: PixelLayer):
    base = PSDImage.open(full_name(filename))
    if filename == "artboard.psd":
        base[0].append(pixel_layer)
    else:
        base.append(pixel_layer)
    with tempfile.TemporaryDirectory() as tmpdir:
        base.save(tmpdir + "/base_file_with_pixel_layer.psd")


@pytest.mark.parametrize(
    "filename",
    [
        "1layer.psd",
        "2layers.psd",
        "artboard.psd",
    ],
)
def test_add_pixel_layer_from_pil_and_write(filename: str):
    import PIL

    base = PSDImage.open(full_name(filename))
    pil_rgb = PIL.Image.new("RGB", (30, 30))
[...]

PR #438, opened 2024-10-08
Artboard lacking the bounding record/channels fix.

al3ks1s:
Fix for issue https://github.com/psd-tools/psd-tools/issues/437.

The artboard _move function will set the _bounding_records, _bounding_channels.


kyamagu:
@al3ks1s Thanks! Can you prepare a test case to check this?

PR #457, opened 2025-01-19
Force image data update on saving

al3ks1s:
Updates the ImageData section when saving the psd if the layer structure was updated.


kyamagu:
Thanks for this feature.
Probably, this should be explicitly documented in 1) the docstring of the `save` method and 2) the general usage document.

kyamagu (approved):
LGTM

PR #628, opened 2026-04-11
feat: Add adjustment layers to composite pipeline

chucorate:
## Summary

- Adjustment layer support was added to the current composite pipeline. 
- New adjustment composite functions were added:
  1. Brightness & Contrast (RGB, CMYK, Grayscale)
  2. Levels (RGB, CMYK, Grayscale)
  3. Curves (RGB, CMYK, Grayscale)
  4. Exposure (RGB, Grayscale)
  5. Invert (RGB, CMYK, Grayscale)
  6. Posterize (RGB, CMYK, Grayscale)
  7. Threshold (RGB, Grayscale)
- The composite of pass-through groups was changed in order to reproduce faithfully the behaviour of pass-through adjustment layers.

The composing quality of each adjustment can be checked in the following folder, that has videos of all algorithms in action.

[Showcases](https://drive.google.com/drive/folders/1Uc4CcBwzqTYVRtQ84yifrFKyBbKjDB3A?usp=sharing)

**Note:** Documentation hasn't been created yet, I want to make sure all changes are reviewed first.

### composite/composite.py

A new function `_apply_adjustment(...)` was added, which composes the adjustment layer on the current backdrop. The decision of making a new function instead of incorporating the behaviour into an existing one was taken to differenciate adjustments as layers that don't add new pixels, but only modify already existing ones. This function calls the `ADJUSTMENT_FUNC` dictionary located in `psd_tools.composite.adjustments`, that will map an adjustment layer to its corresponding composing function. This mimics the current implementation of blend modes.

`_apply_source(...)` has been split into `_apply_source(...)` and `_apply_passthrough_source(...)`, where the first function handles "over" composing of layers the same way as before, while the second function handles specifically the composing of passthrough layers. This change was done to isolate the logic of pass-through composing, as adjustment layers can pass-through the backdrop, but don't add new pixels in it. In this sense, the result of the composite of a pass-through group will blend with the backdrop using linear interpolation, instead of getting composed over it.

The `adjustment_isolated` flag was added to `Compositor`, in order to isolate adjustment layers on the inmediate group they are contained with, even if the group has the pass-through blendmode. This is used in the case the group has its fill attribute not set to 100, where pixel layers still can pass through to the backdrop colors, but adjustments cannot. This was tested in the `tests/psd_files/passthrough_fill_adjustment.psd` file.

### composite/adjustments.py

A new file `adjustments.py` was added to the `composite` module, which has the implementations of all current adjustment composition functions. Each function takes the following parameters as arguments:
- `img: np.ndarray` - The image to apply the adjustment.
- `colormode: Literal[ColorMode.CMYK, ColorMode.GRAYSCALE, ColorMode.RGB]` - The color space in which to apply the adjustment.
- `layer: AdjustmentLayer` - The layer that stores the adjustment's parameters.
All functions will output a `np.ndarray` array, which is the image with the adjustment applied.

As the way each adjustment stores its parameters in `api/adjustments` is different, the adjustment composition functions take the *layer* in order to read the parameters, instead of a unified descriptor-like object. This is an aspect that can be improved upon discussion, but as a starting point I think the current implementation is valid.  

## Other Changes

- `api/pil_io.py: _apply_icc(...)` - This function was changed in order to handle without errors the application of ICC profiles on images with alpha channels, which was an issue that I stumbled upon testing composing of grayscale images. Now, if an image has an alpha channel, it will be temporarily separated from the color channels, then the ICC profile gets applied to the color channels, and finally the alpha channel gets attached to the resulting RGB image.
- `composite/composite.py: _is_passthrough(...)` - This function was added to check if a group layer will indeed be applied as a passthrough group. This was done because pass-through groups that have clipping layers will use normal blending, even if their blending mode is set to pass-through.
- `composite/composite.py: _get_group(...)` - If a layer is a pass-through group, the whole viewport will be passed on to the group compositor, so adjustment layers affect all the backdrop.
- `composite/composite.py: _get_mask(...)`- Fixed incorrect handling of vector mask density when it was set to a value different from 255. Now `vector_mask_density` gets applied as a blend between the vector mask and the full coverage. This is tested in the `tests/psd_files/passthrough_vector_mask.psd` file.

## Test Coverage
The following tests where added:
- Tests for each adjustment composite function, for each color mode. In each test, the adjustment gets applied four times (disjointed from each other) with different parameters. Additionally, for each adjustment there's a color palette to study the composite fidelity to Photoshop's implementation.
- Tests for nested group composition of layers were added, in order to catch the behaviour of layers when combinations of pass-through and non-pass-through groups are applied. 
- Tests for adjustment clipping, as well as controlled pass-through groups under different opacity, fill, clipping, masks and blendmode settings were added, to test the fidelity of pass-through composing.

This adds up to 35 new test psd files. At the time of this pull request, all 1265 tests pass upon running `uv run pytest`.


chucorate:
I didn't know making a commit would stop the copilot review, I'm sorry :/. I won't make any changes until everything is done

chucorate:
It failed again. I'll fix the remaining format and typing issues in the meantime

EDIT: I ran `uv run ruff format --check` and `uv run mypy` sucessfully, tests should work now

kyamagu:
Thanks for the PR. This will be a significant functional change and requires careful review. Let me first ask Claude to review

kyamagu (changes requested):
## Code Review

Thanks for the contribution — this is a well-structured feature. The dispatch-table pattern (`ADJUSTMENT_FUNC`) mirrors the existing blend-mode design cleanly, the test coverage is thorough (35 new fixtures across all colour modes and compositing scenarios), and the pass-through refactoring is a real improvement. I'm requesting changes for the confirmed bugs below before merge. See inline comments for details.

### Confirmed bugs that need fixing

| # | Severity | Location | Summary |
|---|---|---|---|
| 1 | High | `adjustments.py` `_apply_lut` | Off-by-one: `depth = lut_size` should be `lut_size - 1` |
| 2 | High | `adjustments.py` `apply_posterize` | Divide-by-zero when `levels == 1` |
| 3 | High | `composite.py` `_get_mask` | Density refactor reverses original fallback logic, discarding `user_mask_density` |
| 4 | High | `adjustments.py` `apply_exposure` | Exposure exponent divided by `color_gamma` — not part of Photoshop's documented formula |
| 5 | Medium | `composite.py` `_apply_luts` | Master LUT applied *after* per-channel LUTs; Photoshop applies master *first* |
| 6 | Medium | `composite.py` `_is_passthrough` | Conflates clip-layer presence with pass-through semantics |

### Other concerns

- **Brightness/Contrast**: the TODO "might be slightly off" should be replaced by either a verified algorithm or tests that enforce the allowed deviation.
- **Threshold `- 0.01`**: unexplained magic constant needs a comment.
- **`apply_invert`**: operates on the colour-only buffer so it is fine now, but an assertion or comment would prevent silent alpha inversion if the function is reused later.
- **`if alpha:`** in `pil_io.py`: not a crash (PIL Images are truthy), but `if alpha is not None:` matches the intent.
- Missing test: adjustment layer inside a pass-through group *that also has clip layers* (exercises the `_is_passthrough` edge case).

src/psd_tools/composite/adjustments.py
```diff
@@ -0,0 +1,298 @@
...
+    from scipy.interpolate import CubicSpline  # type: ignore[import-untyped]
+
+    use_legacy: bool = layer.use_legacy
+    b: float = layer.brightness / 150.0
+    c: float = layer.contrast / 100.0
+
+    lut_size = _get_lut_size(layer)
+    t = np.linspace(0, 1, lut_size, dtype=np.float32)
+
+    if use_legacy:  # these layers are skipped during composing as they are recognized as PixelLayers with no bounding box
```
kyamagu, line 93: **Bug: off-by-one quantization**

`depth = lut_size` creates `N` buckets of width `1/N`, but a 256-entry LUT maps indices `0..255`, so the correct scale factor is `lut_size - 1`. With the current code all quantized values are shifted slightly high.

```python
# Fix:
depth = lut_size - 1
values = (np.floor(values * depth) / depth).clip(0.0, 1.0)
```

src/psd_tools/composite/adjustments.py
```diff
@@ -0,0 +1,298 @@
...
+    colormode: Literal[ColorMode.CMYK, ColorMode.GRAYSCALE, ColorMode.RGB],
+    layer: Posterize,
+) -> np.ndarray:
+
+    lut_size = _get_lut_size(layer)
+    levels = layer.posterize
+
+    values = np.linspace(0.0, 1.0, lut_size, dtype=np.float32)
+
+    lut = (np.floor(levels * values) / (levels - 1)).astype(np.float32)
```
kyamagu, line 258: **Bug: divide by zero when `levels == 1`**

Photoshop allows posterize level 1. `levels - 1 == 0` produces a `ZeroDivisionError` or NaN array. Add a guard:

```python
if levels <= 1:
    return np.zeros_like(img)
lut = (np.floor(levels * values) / (levels - 1)).astype(np.float32)
```

src/psd_tools/composite/adjustments.py
```diff
@@ -0,0 +1,298 @@
...
+
+    exposure: float = layer.exposure
+    offset: float = layer.exposure_offset
+    gamma: float = layer.gamma
+
+    lut_size = _get_lut_size(layer)
+    color_gamma = 1.8 if colormode == ColorMode.GRAYSCALE else 2.2
+    values = np.linspace(0.0, 1.0, lut_size, dtype=np.float32)
+
+    factor = np.pow(2.0, exposure / color_gamma)
```
kyamagu, line 224: **Bug: exposure formula divides exponent by `color_gamma`**

Photoshop's documented Exposure formula is:



Dividing `exposure` by `color_gamma` (1.8/2.2) incorrectly ties the exposure strength to the display gamma. `color_gamma` is not part of Photoshop's Exposure math and should not appear in the exponent.

Suggested fix:

src/psd_tools/composite/adjustments.py
```diff
@@ -0,0 +1,298 @@
...
+    channels = range(1, n_channels + 1)
+
+    # individual adjustments get applied independently of each other, then the master lut its applied if exists
+    for channel_id in channels:
+        if channel_id in luts:
+            lut = luts[channel_id]
+            out[:, :, channel_id - 1] = _apply_lut(img[:, :, channel_id - 1], lut)
+
+    if not colormode == ColorMode.GRAYSCALE and 0 in luts:
+        out = _apply_lut(out, luts[0])
```
kyamagu, line 38: **LUT application order appears reversed**

Photoshop applies the master channel LUT first (as a global input remap), then per-channel LUTs refine from that remapped result. This code applies per-channel LUTs first and then the master — the order is backwards and will produce wrong output for non-symmetric curves/levels.

src/psd_tools/composite/composite.py
```diff
@@ -253,6 +254,11 @@ def _blend_backdrop(
     return result_color, result_alpha
 
 
+def _is_passthrough(group: Layer) -> bool:
+    """Check if group layer will be composed as a pass-through group."""
```
kyamagu, line 258: **Pass-through semantics conflated with clip-layer presence**

A group having clip layers is a structural property; it should not disable pass-through compositing. With this logic, any pass-through group that happens to have clipping layers will silently be treated as an isolated group, changing blending behaviour without any indication.

Pass-through mode should be determined solely from the blend mode (and the isolated flag if relevant), with clip-layer handling kept as a separate compositing step.

src/psd_tools/composite/composite.py
```diff
@@ -518,10 +614,11 @@ def _get_mask(self, layer: Layer) -> tuple[float | np.ndarray, float]:
                 )
             if layer.mask.parameters:
                 density = layer.mask.parameters.user_mask_density
-                if density is None:
-                    density = layer.mask.parameters.vector_mask_density
-                if density is None:
-                    density = 255
+                density = (
+                    layer.mask.parameters.vector_mask_density
```
kyamagu, line 618: **Bug: density refactor reverses the original fallback logic**

The original code was:
```python
density = layer.mask.parameters.user_mask_density
if density is None:
    density = layer.mask.parameters.vector_mask_density
if density is None:
    density = 255
```

The new code, when `user_mask_density is not None`, replaces it with `vector_mask_density` — the opposite of the intended fallback. And when `user_mask_density is None`, it jumps straight to `255`, ignoring `vector_mask_density` entirely.

`user_mask_density` is effectively discarded.

src/psd_tools/api/pil_io.py
```diff
@@ -327,6 +338,10 @@ def _apply_icc(image: Image.Image, icc_profile: bytes) -> Image.Image:
         logger.error("Failed to apply ICC profile.")
         return image
 
+    if alpha:
```
kyamagu, line 341: Nit: use `if alpha is not None:` — PIL Images do not raise on `bool()` so this is not a crash, but `is not None` matches the intent (checking whether a channel was extracted) more clearly.

src/psd_tools/composite/adjustments.py
chucorate: The implementation of exposure is based on [this article](https://geraldbakker.nl/psnumbers/exposure.html), which has the best explanation of the exposure algorithm I could find, and *does* consider the gamma value of the ICC profile. It's pretty accurate, as the MSE difference between the outputs produced in tests and the reference images is less than 0.0005, but as seen from the [exposure video I sent](https://drive.google.com/file/d/1uLEHU1C4NB_6wmQMvXa_Tiysy32UrxPe/view?usp=drive_link) (timestamp 0:53 onwards), the algorithm may sometimes produce slightly darker images.  

Ignoring `color_gamma` (making it equal to 1) will yield completely inaccurate results and fail all exposure related tests. Given this, there's value in using `color_gamma`, but I could try improving the algorithm to make it more accurate in those situations where the output is slightly darker than the reference. Alternatively, I can leave this code as it is, and mark it as a *TODO* for future improvement, since the current implementation already works fine and the darker outputs are not the norm.  

Unfortunately, most adjustment algorithms (including exposure) don't have official photoshop documentation available, or at least I couldn't find any. What do you think is the best course of action to take here?

src/psd_tools/composite/adjustments.py
chucorate: The current order (master last) is correct. Here's a [proof](https://drive.google.com/file/d/1s8fGBYnAIXNm3pXg-gd-ekFqE_C05zgd/view?usp=drive_link). The experiment first registers the result of applying all luts in a single layer, and then uses two groups that decompose the luts in individual adjustments. The one were the master lut is applied last matches exactly with the reference, as seen by the difference and subtract blending modes. On the other hand, the one were the master lut is applied first is not equal to the reference. The adjustments were tested under arbitrary curves, so no symmetry could alter the results.

On the code's side, changing the lut order in `adjustments.py` will make most adjustment tests fail.

src/psd_tools/composite/composite.py
chucorate: This is true, and it's something I didn't catch. Just like when changing the fill of a pass-through group to a non 100 value, having clipping layers on a pass-through group will make *pixel* layers pass but **not** adjustment layers, as seen on [this video](https://drive.google.com/file/d/1gGplAz8YJlpN8FrTy-gIilJgWv7XR7ND/view?usp=drive_link). In this sense, the solution would be to process this discrepancy using the `adjustment_isolated` flag, and changing back the `is_passthrough` condition to check only for the blendmode.

This will be changed, and new tests for this situation will be added.

chucorate:
Thanks for the review! I will try to address all the requests in the following days/weeks, as most changes break one or more tests

src/psd_tools/composite/composite.py
chucorate: The original logic was added back. Additionally, The way the density on usermasks is managed was partially fixed, as density was previously reducing the opacity *inside* the mask, instead of making areas outside the mask more visible.

```py
if layer.mask.parameters:
    density = layer.mask.parameters.user_mask_density
    if density is None:
        density = layer.mask.parameters.vector_mask_density
    if density is None:
        density = 255

    density = float(density) / 255.0
    shape = density * shape + (1 - density)
```

To test this, the following tests were added:
- `tests\psd_files\mask-density-layermask.psd`
- `tests\psd_files\mask-density-layervectormask.psd` 
- `tests\psd_files\mask-density-vectormask.psd`

`tests\psd_files\mask-density-layervectormask.psd` was marked as xfail, because usermask currently mixes layer and vector mask properties. When applying density in isolation tests pass. All other tests that had masks set to different densities pass as well.

As this PR is focused on adjustments & passthrough composing, I think it could be a good idea to fix this on its entirety on a different PR

src/psd_tools/composite/composite.py
chucorate: This was fixed on the following commit: https://github.com/psd-tools/psd-tools/pull/628/changes/2aa5e31d8845d696efb71df400d1ab23ef364de4

When a group is passthrough, but adjustments are *not* passed through the backdrop, the composing logic falls back to `apply_source`. This can be seen mainly on the following lines:

In `self.apply()`:
```py
full_passthrough = (
    layer.blend_mode == BlendMode.PASS_THROUGH
    and is_adjustment_isolated is False
)  # when adjustments are isolated, passthrough composing fallbacks to over composing

if full_passthrough:
    self._apply_passthrough_source(
        color, shape * shape_const, alpha * shape_const, mask * shape_const
    )
else:
    self._apply_source(
        color,
        shape * shape_const,
        alpha * shape_const,
        layer.blend_mode,
        knockout,
    )
```

In  `self._get_group()`:
```py
if isolate_adjustments:
    color = group_compositor.color  # prevents backdrop color contamination
else:
    color = group_compositor._color
```

`_get_group()` also returns the `isolate_adjustments` flag, so the group can be composed properly on `apply`.

More passthrough related tests were added (see commit). All passthrough related tests pass.

src/psd_tools/composite/adjustments.py
chucorate: This won't be changed because this order is correct, as seen from the test results.

Tests in `test_adjustments.py` have been made stricter overall, and the file's structure was refactored so different tests could be applied in a more readable way to specific adjustments. See commit https://github.com/psd-tools/psd-tools/pull/628/changes/db5fb0513399236c4442e960a2129e16acfe726d

This change was also made to test adjustments that, while they work, their accuracy can be improved, like the brightness/contrast adjustment algorithm

src/psd_tools/composite/adjustments.py
chucorate: `color_gamma` is applied separately now. Additionally, removing clipping when exposure is applied resolved the brightness issues. 

```py
lut = np.power(values, color_gamma)  # image is converted to linear color space first

# exposure operates in linear space
lut *= np.exp2(exposure)
lut = (lut + offset).clip(0.0, 1.0)
lut **= 1.0 / gamma

lut **= 1.0 / color_gamma  # convert back to original space using inverse TRC approximation
lut = lut.clip(0.0, 1.0).astype(np.float32)
```

See commit https://github.com/psd-tools/psd-tools/pull/628/changes/1f3533aeb65dfed76ed08e8f56fa77d9512d5afa for more details.

src/psd_tools/api/pil_io.py
chucorate: This was fixed in the following commit: https://github.com/psd-tools/psd-tools/pull/628/changes/1f3533aeb65dfed76ed08e8f56fa77d9512d5afa

src/psd_tools/composite/adjustments.py
chucorate: Photoshop doesn't allow level values outside the [2, 255] range on posterize. The code was modified to reflect that.

```py
lut_size = _get_lut_size(layer)
levels = max(2, min(layer.posterize, 255))
correction_factor = (
    (lut_size - 1) / lut_size
)  # including this factor makes the output more accurate, on 8 bits and 16 bits

out = np.floor(correction_factor * img * (levels)) / (levels - 1)
```

See commit https://github.com/psd-tools/psd-tools/pull/628/changes/1f3533aeb65dfed76ed08e8f56fa77d9512d5afa for more details.

src/psd_tools/composite/adjustments.py
chucorate: This was fixed on commit https://github.com/psd-tools/psd-tools/commit/1f3533aeb65dfed76ed08e8f56fa77d9512d5afa. 

Additionally, correction constants were added to the `posterize` and `threshold` adjustment algorithms, in order to preserve (and increase) their accuracy. 

On posterize:
```py
lut_size = _get_lut_size(layer)
levels = max(2, min(layer.posterize, 255))
correction_factor = (
    (lut_size - 1) / lut_size
)  # including this factor makes the output more accurate, on 8 bits and 16 bits

out = np.floor(correction_factor * img * (levels)) / (levels - 1)
```

On threshold:
```py
lut_size = _get_lut_size(layer)

correction_factor = (lut_size - 1) / lut_size
bit_factor = (lut_size - 1) / 255  # constant used to compensate values on 16 bits
bit_offset = 1 / 255  # offset used to compensate values on 16 bits
threshold = (layer.threshold - bit_offset) * correction_factor * bit_factor

luminance = np.round(_get_luminance(img, colormode) * (lut_size - 1))
filtered = (luminance > threshold).astype(np.float32)
```

To test this, new posterize test files were added for 16bit composing, and tests were made stricter.

chucorate:
To address the `apply_invert` comment, a new decorator `_preserve_alpha` was added to `adjustments.py`. With this, all adjustments are applied only on color channels, leaving the alpha channel intact if it exists.

If all current changes are okay, should I get to work on adding documentation?

src/psd_tools/composite/adjustments.py
kyamagu: Ok

src/psd_tools/composite/adjustments.py
kyamagu: Ok, leaving a TODO comment looks good at this point. Thanks!

kyamagu (approved):
LGTM

kyamagu:
There is a CPython 3.10 incompatibility.

kyamagu:
Actually, the compositing subpackage has had bad typing safety. Refactoring this package would be future work

kyamagu:
@chucorate Thanks for the big effort!

chucorate:
Sure! I'm planning on adding more adjustments in the future :)

PR #647, opened 2026-05-04
Use `IO[bytes]` instead of `BinaryIO`

tyilo:
`IO[bytes]` is less restrictive.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -124,10 +124,10 @@ All binary structures inherit from `BaseElement` and implement:
 
 ```python
 @classmethod
-def read(cls, fp: BinaryIO, **kwargs) -> Self:
+def read(cls, fp: IO[bytes], **kwargs) -> Self:
     """Read from file pointer"""
 
-def write(self, fp: BinaryIO, **kwargs) -> int:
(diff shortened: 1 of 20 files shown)


kyamagu:
@tyilo What is the use case you want to address with `IO[bytes]`?

tyilo:
> @tyilo What is the use case you want to address with `IO[bytes]`?

`IO[bytes]` is less restrictive than `BinaryIO` and `psd-tools` never uses any function from `BinaryIO` (the only one being `__enter__`). I only have an `IO[bytes]` instance not a `BinaryIO` instance and want to process it with `psd-tools`.

There is a related issue here: https://github.com/cgohlke/tifffile/issues/268

For instance, `PIL.Image.open` takes `IO[bytes]` and not `BinaryIO` as then it can be used in more cases.

@kyamagu Is there a specific reason why you want to use `BinaryIO` and not just `IO[bytes]` when it gives the caller more flexibility?

kyamagu:
Understood. `IO[bytes]` is less readable than `BinaryIO`, IMO, but it is absolutely fine to change the type to `IO[bytes]` for use cases like zip files.

kyamagu (approved):

Issue #382, opened 2024-01-03
Outer Glow not being composited


bertou:
A simple Outer Glow is properly reproduced by psd.composite().save() but not by psd.composite(force=True).save()

Without forcing composite:

<img width="284" alt="Screenshot 2024-01-03 at 17 50 50" src="https://github.com/psd-tools/psd-tools/assets/11999419/c024b790-0602-4acf-bcab-625430d6a45d">

With forcing:

<img width="282" alt="Screenshot 2024-01-03 at 17 51 38" src="https://github.com/psd-tools/psd-tools/assets/11999419/4bb99469-81e2-4a7b-a6a6-45b2be7a0344">

kyamagu:
This feature is not supported yet

bertou:
@omar16100 code fixes the issue, but doesn't handle the element part which is doing some extra blurring...

<img width="364" height="100" alt="Image" src="https://github.com/user-attachments/assets/0bad30b0-f46b-41d9-a2a9-320245123f7c" />

Issue #326, opened 2022-07-05
Shadow on psd thumbnail


andrey-hider:
Hi,
I have a psd file where shadow was defined by two layer masks.

**To Reproduce**
```
from PIL import Image

folder = "/Users/admin/Python/issue/shadow_psd_tools/"
file = "example.psd"

from psd_tools import PSDImage
psd = PSDImage.open(folder + "/" + file)
psd.composite().save(folder + '/psd_tools.png')
```
**Environment**

- MacOS [12.4]
- Python [3.9.13]
- psd-tools [1.9.21]

Thumbnail that I got out of selected code:
![psd_tools](https://user-images.githubusercontent.com/30869869/177306129-45d2c45c-9d2d-449d-b5f9-b3992a4e7944.png)

This is how preview looks in Photoshop:
<img width="1546" alt="Screenshot 2022-07-05 at 12 08 50" src="https://user-images.githubusercontent.com/30869869/177306251-e1e7f086-9661-4604-94c3-1f4c1da854be.png">

Is there any workaround to avoid this issue?
I see that with the disabled layer "Farbfüllung 1" psd-tools is generating correct thumbnail.
Added archive with the code and psd file: [shadow_psd_tools.zip](https://github.com/psd-tools/psd-tools/files/9045497/shadow_psd_tools.zip)

PR #679, opened 2026-06-24
feat: render drop shadow layer effect

omar16100:
## Summary

`composite()` already parses every drop-shadow parameter (`layer.effects` →
`DropShadow.distance/.angle/.size/.choke/.color/.blend_mode/.layer_knocks_out`) but never
rendered the effect — the shadow silently disappears on `composite(force=True)`. This adds drop
shadow rendering to the compositing pipeline.

Relates to #326 (shadow missing from thumbnail/composite). Outer glow (#382) and inner shadow/glow
are natural follow-ups using the same machinery.

## Approach

- **`composite/effects.py`** — new `draw_drop_shadow(viewport, shape, *, distance, angle, size, choke)`
  that synthesizes a coverage mask from the layer silhouette: choke (morphological erosion, as a
  percent of size), gaussian blur, then offset opposite the light direction. Mirrors
  `draw_stroke_effect`'s `@require_scipy` gate and local-import convention.
- **`composite/composite.py`** — `Compositor._apply_drop_shadow()` is called in `apply()` just
  *before* the layer's own source is composited, so the shadow paints behind the layer onto the real
  backdrop. It reuses the existing primitives rather than reinventing them:
  - `paint.draw_solid_color_fill(..., effect.value)` for the tint (handles every color space),
  - `self._apply_source(..., effect.blend_mode)` for blending (the bytes `Enum` matches `BLEND_FUNC`
    by value, exactly like the overlay effects),
  - the `_get_stroke` bbox-expansion idiom to grow the working viewport for the blur/offset.
  - Honors `effect.layer_knocks_out` (the shadow is concealed under the layer's own coverage) and
    `layer.effects.find("dropshadow")` (master switch + per-effect enable).

The module docstring promotes drop shadow from "limited" to "supported".

## Fidelity (honest)

Placement is faithful: against `tests/psd_files/layer_effects.psd`, the rendered shadow correlates
**~0.93** with Photoshop's own baked darkening and the shadow-zone MSE vs the embedded preview drops
sharply. However, the falloff is a **gaussian approximation of Photoshop's proprietary blur**, so
shadow *intensity* can differ — large, soft shadows tend to render somewhat darker than Photoshop
(Photoshop's falloff is simultaneously concentrated and low-peak, which a single gaussian/box blur of
a silhouette can't reproduce). This is documented in the module docstring. I'd rather ship faithful
placement with an honest caveat than nothing; happy to iterate on the falloff if you have a preferred
model.

## Tests

- Unit tests for `draw_drop_shadow` (offset direction, blur softening/spread, choke tightening).
- An integration test on `layer_effects.psd` asserting the shadow moves the `force=True` composite
  closer to Photoshop's baked preview than a no-op renderer does (falsifiable — a zero renderer ties).
- Full `tests/psd_tools/composite/` suite passes (195 passed); `ruff`, `ruff format`, and `mypy` clean.

## Open questions

- Happy to gate the new render test behind `xfail` (like the stroke render tests) if you prefer a
  strict pixel threshold over the relative-improvement assertion.
- Scope check: would you welcome outer glow / inner shadow as follow-up PRs on the same pattern?

### DIFF
--- a/src/psd_tools/composite/composite.py
+++ b/src/psd_tools/composite/composite.py
@@ -1,6 +1,7 @@
 """Composite implementation for layer rendering and blending."""
 
 import logging
+import math
 from typing import Callable, cast
 
 import numpy as np
@@ -13,7 +14,7 @@ from psd_tools.api.utils import EXPECTED_CHANNELS, check_pixel_size
 from psd_tools.composite import paint, utils, vector
 from psd_tools.composite.adjustments import ADJUSTMENT_FUNC
 from psd_tools.composite.blend import BLEND_FUNC, normal
-from psd_tools.composite.effects import draw_stroke_effect
+from psd_tools.composite.effects import draw_drop_shadow, draw_stroke_effect
 from psd_tools.constants import BlendMode, ColorMode, Resource, Tag
 
 logger = logging.getLogger(__name__)
@@ -353,6 +354,9 @@ class Compositor(object):
         alpha *= mask
 
         # TODO: Tag.BLEND_INTERIOR_ELEMENTS controls how inner effects apply.
+
+        # Outer effects paint behind the layer's own source.
+        self._apply_drop_shadow(layer, color, shape, alpha)
 
         full_passthrough = (
             layer.blend_mode == BlendMode.PASS_THROUGH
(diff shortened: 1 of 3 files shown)


kyamagu:
@omar16100 Thanks. Can you at least address Copilot's suggestions? I have a concern about not having test fixtures to measure fidelity here

bertou:
I tested it (as I always wanted to have outer_glow working, #382 ), and it works fine **with** copilot corrections which are indeed correct (at least the 2 relative to adding the layer opacity).
It does not fully solve my issue in #382 as it doesn't take into account the "element" part of the effect, which in my case is used (softer instead of precise, with some values for spread and size). If I turn my element to precise I get the correct rendering. I guess for someone who understands what it does (or to copilot?) it may be straightforward to implement?
In my case it would be night and day as I could use psd-tools to generate the png instead of using it only to parse the psd files and then dump a photoshop jsx to process the files and export pngs...

PR #680, opened 2026-06-24
feat: render outer glow layer effect

### LINKED ISSUE #382: Outer Glow not being composited

omar16100:
## Summary

Adds **Outer Glow** rendering to `composite()`, the natural follow-up to drop shadow. `composite()`
already parses `OuterGlow` parameters but never rendered them, so the glow disappears on
`composite(force=True)`.

**Closes #382.**

> ⚠️ **Stacked on #679** (drop shadow). This branch is based on the drop-shadow branch, so until #679
> merges this PR's diff shows *both* commits. The net-new change here is the single commit
> **"feat: render outer glow layer effect"** — please review #679 first. Once #679 lands I'll rebase and
> the diff will reduce to just the glow change.

## Approach

Mirrors the drop-shadow renderer with three deltas that define an outer glow:

- **`composite/effects.py`** — new `draw_outer_glow(viewport, shape, *, size, spread)`: **dilates** the
  silhouette by spread (vs. a shadow's erode-choke), gaussian-blurs by size, and applies **no offset**
  (the glow is centered).
- **`composite/composite.py`** — `Compositor._apply_outer_glow()`, called right after
  `_apply_drop_shadow` so the glow paints *behind* the layer, then **unconditionally knocked out under
  the layer's own coverage** (`mask *= 1 - shape`), since an outer glow is by definition outside the
  layer. Reuses `_apply_source` + `BLEND_FUNC` (Screen) and `paint.draw_solid_color_fill`.

Solid color only; gradient glow (`OuterGlow.gradient`) is deferred. The module docstring promotes
outer glow to "supported".

## A heads-up on `OuterGlow.spread`

The renderer reads the spread amount from **`effect.choke`** (the `ChokeMatte` / `b'Ckmt'` descriptor
slot — where Photoshop actually stores a glow's "Spread"), **not** from `effect.spread`. The
`OuterGlow.spread` accessor currently reads `Key.ShadingNoise` (`b'ShdN'`, the *jitter* control), which
looks like a latent bug. I left the accessor alone to keep this PR focused — happy to send a separate
fix if you agree.

## Fidelity

Same honest caveat as drop shadow: placement is faithful (against `layer_effects.psd` the rendered
glow moves the `force=True` composite markedly closer to Photoshop's baked preview), but the falloff is
a gaussian approximation of Photoshop's proprietary blur, so intensity can differ.

## Tests

- Unit tests for `draw_outer_glow` (symmetric/no-offset spread; spread dilates the footprint).
- An integration test on `layer_effects.psd` asserting the glow moves the `force=True` composite closer
  to the baked preview than a no-op renderer (falsifiable — a zero renderer ties).
- Full `tests/psd_tools/composite/` suite passes (198); `ruff`, `ruff format`, `mypy` clean.

### DIFF
--- a/src/psd_tools/composite/composite.py
+++ b/src/psd_tools/composite/composite.py
@@ -1,6 +1,7 @@
 """Composite implementation for layer rendering and blending."""
 
 import logging
+import math
 from typing import Callable, cast
 
 import numpy as np
@@ -13,7 +14,7 @@ from psd_tools.api.utils import EXPECTED_CHANNELS, check_pixel_size
 from psd_tools.composite import paint, utils, vector
 from psd_tools.composite.adjustments import ADJUSTMENT_FUNC
 from psd_tools.composite.blend import BLEND_FUNC, normal
-from psd_tools.composite.effects import draw_stroke_effect
+from psd_tools.composite.effects import draw_drop_shadow, draw_stroke_effect
 from psd_tools.constants import BlendMode, ColorMode, Resource, Tag
 
 logger = logging.getLogger(__name__)
@@ -353,6 +354,9 @@ class Compositor(object):
         alpha *= mask
 
         # TODO: Tag.BLEND_INTERIOR_ELEMENTS controls how inner effects apply.
+
+        # Outer effects paint behind the layer's own source.
+        self._apply_drop_shadow(layer, color, shape, alpha)
 
         full_passthrough = (
             layer.blend_mode == BlendMode.PASS_THROUGH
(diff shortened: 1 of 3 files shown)


kyamagu:
@omar16100 Thanks for the PR. Can you at least address Copilot's suggestions, regardless of whether you accept or reject the changes?

Issue #645, opened 2026-04-29
Files with Hue/Saturation adjustments don't pass read-write tests


chucorate:
**Describe the bug**
When a file has Hue/Saturation adjustments, it won't pass the `test_psd_read_write` test at `tests\psd_tools\psd\test_psd.py`. In particular, the output will have $36\cdot n$ less bytes than the expected result, where $n$ is the amount of Hue/Saturation adjustments present in the file. The discrepancy is related to *layer and mask* info, as seen from the screenshot below.

**To Reproduce**
Download the psd file below and execute the following code (it replicates the `test_psd_read_write` test):

```python
import io
import logging

from psd_tools.psd import PSD

logging.basicConfig(level=logging.DEBUG)


def test_psd_read_write(filename: str) -> None:
    with open(filename, "rb") as f:
        expected = f.read()

    with io.BytesIO(expected) as f:
        psd = PSD.read(f)

    with io.BytesIO() as f:
        psd.write(f, padding=4)
        f.flush()
        output = f.getvalue()

    print(len(output), len(expected))
    print(output == expected)

test_psd_read_write("huesaturation read-write.psd")
```

**Expected behavior**
The output should exactly equal the expected result.

**File and screenshots**
File used: https://drive.google.com/file/d/1OHVAIjQGg_scZ_m35lUGOXVCnbwZjDyq/view?usp=sharing
Console screenshot:

<img width="1058" height="424" alt="Image" src="https://github.com/user-attachments/assets/b7348035-53ea-45a6-b5cb-10375b72b1ae" />

**Environment**
 - Python [3.13.5]
 - psd-tools version [v1.16.0]
 - Photoshop [27.1.0]

kyamagu:
@chucorate This can happen because byte-by-byte correspondence is not required as per the PSD spec. Third-party authoring tools often produce different byte streams than Photoshop. Write-read round tests have more importance

chucorate:
In that case, should I add files with hue/saturation adjustments to `SKIP_BYTE_ROUND_TRIP` so they are not used in `test_psd_read_write`? 

For context, I'm adding the Hue/Saturation composite algorithm, but read-write tests fail. All other tests (including write-read) tests do pass. I can just submit the pull request with failing tests if further context is needed

kyamagu:
@chucorate Yes, you may add fixtures to `SKIP_BYTE_ROUND_TRIP`. Just make sure the written files can be opened in authoring tools without an issue.

PR #794, opened 2026-09-06
fix: keep the bytes after a Hue/Saturation block's range records (#645)

MaxFreedomPollard:
Saving a file whose Hue/Saturation layers came from a recent Photoshop drops 36 bytes out of each `hue2` block. The block Photoshop writes is 136 bytes; psd-tools writes 100.

`HueSaturation.read()` at `src/psd_tools/psd/adjustments.py:515` reads exactly the 100 bytes the format documents: version and colorization flag, the colorization triple, the master triple, and six range records of four range values plus three settings. It never touches the rest of the stream, and `BaseElement.frombytes()` at `src/psd_tools/psd/base.py:73` hands `read()` a `BytesIO` over the whole block and throws away what is left in it, so the trailing bytes are gone by the time `write()` at `src/psd_tools/psd/adjustments.py:533` runs, and the block comes back out 36 bytes shorter.

Those 36 bytes are six `(hue, 100, 50)` triples, one per range, and they are not constant. Across the 16 Hue/Saturation blocks in `tests/psd_files/adjustments/` each triple's hue is the midpoint of that range's inner band, which moves when the user drags the range sliders: in `huesaturation_saturation_rgb.psd` one layer's first range is `(103, 111, 157, 165)` and its first trailing triple is `(134, 100, 50)`. That is why #645 came out 36 bytes per layer short, and it is data rather than the padding difference the issue was closed as.

The fix adds an `unknown: bytes` field that captures the remainder and writes it back, the same way `ChannelMixer` two structures above in the same file already handles its own trailer. Files from older Photoshop stop at 100 bytes and leave the field empty, so nothing about them changes.

To check that no other block was doing this, I monkeypatched `TaggedBlock.read` to re-encode every block it parsed and compare against the raw bytes, then ran it over all 295 fixture files. Before the change: 16 `HUE_SATURATION` blocks shrinking 136 to 100, plus 4 `UNICODE_LAYER_NAME` blocks growing by 2 bytes, which is the known non-Adobe padding already listed in `SKIP_BYTE_ROUND_TRIP`. After the change: only the 4 `UNICODE_LAYER_NAME` blocks.

Verification, all on Python 3.11 with `uv sync --group dev --extra composite`:

`uv run pytest --no-cov -q` gives `2890 passed, 3 skipped, 21 xfailed, 2 xpassed` against `2883 passed, 3 skipped, 25 xfailed, 2 xpassed` on `main`. The four fewer xfails are the four huesaturation fixtures, which now reproduce byte for byte, so `test_psd_read_write` no longer needs `HUESATURATION_MISMATCH`.

`uv run pytest --no-cov -q tests/psd_tools/psd/test_adjustments.py -k hue_saturation` gives `3 passed`. Reverting only `src/psd_tools/psd/adjustments.py` and rerunning gives `2 failed, 1 passed`: `test_hue_saturation_rw` with the trailer fails with `136 vs 100`, and `test_hue_saturation_keeps_trailing_bytes` fails with `AttributeError: 'HueSaturation' object has no attribute 'unknown'`. The parametrised case with no trailer passes either way, which is the older 100-byte shape staying put.

`uv run ruff check` gives `All checks passed!`, `uv run ruff format --check .` gives `110 files already formatted`, and `uv run mypy` gives `Success: no issues found in 109 source files`.

Changelog entry added under a new `1.19.1 (unreleased)` heading.


src/psd_tools/psd/adjustments.py
```diff
@@ -537,6 +546,7 @@ def write(self, fp: IO[bytes], **kwargs: Any) -> int:
         for item in self.items:
             written += write_fmt(fp, "4h", *item[0])
             written += write_fmt(fp, "3h", *item[1])
+        written += write_bytes(fp, self.unknown)
         written += write_padding(fp, written, 4)
```
kyamagu: Since `unknown` is `fp.read()` to EOF, it already contains whatever trailing padding the original block had — so padding again after writing it can make the block grow. On this branch a forged 102-byte `hue2` payload round-trips to 104 bytes:

```
100 -> 100 identical
102 -> 104 DIFFERS
103 -> 104 DIFFERS
136 -> 136 identical
```

Dropping the `write_padding` call (the way `ChannelMixer.write` does, which uses the same `unknown: bytes` pattern) makes all four exact:

```suggestion
        written += write_bytes(fp, self.unknown)
```

```
100 -> 100 identical
102 -> 102 identical
103 -> 103 identical
136 -> 136 identical
```

Full suite still passes with that change (2890 passed, 3 skipped, 21 xfailed, 2 xpassed). No file in the corpus hits it — every real block is 100 or 136, both 4-aligned — so this is robustness against non-Adobe producers rather than a regression, but the current ordering inflates the tagged-block length where the old code only truncated it.

kyamagu (commented):
Thanks for this — I checked the claim independently and it holds: decoding every `hue2` block under `tests/psd_files/adjustments/` gives 16 blocks of 136 bytes whose 36-byte tail is six `(hue, 100, 50)` triples, and the hue values genuinely vary per layer (e.g. `huesaturation_saturation_rgb.psd` layer 1 → `134` for inner band `111..157`). So it is data, not padding, and preserving it verbatim is the right call. Full suite on the branch: 2890 passed, 3 skipped, 21 xfailed, 2 xpassed — the four `HUESATURATION_MISMATCH` xfails are legitimately gone.

One small thing about the write path, inline.

kyamagu:
@MaxFreedomPollard Thanks!

Issue #792, opened 2026-09-03
Outside stroke layer effect is clipped to layer.bbox even with an expanded viewport


MisakaCirno:
**Describe the bug**

When compositing a layer that has a Photoshop Stroke layer effect with its
position set to Outside (`OutF`), the part of the stroke outside `layer.bbox`
is clipped.

Passing a manually expanded `viewport` does not solve the problem. It only
adds transparent padding around the original layer bbox; the outside stroke
is still not rendered into that padding.

This can be reproduced with a normal pixel layer and a single solid-color
stroke. Smart objects and multiple strokes are not required.

This may be related to #404, but this report specifically concerns clipping
an outside stroke to `layer.bbox`.

**To Reproduce**

The test layer has these properties:

- RGB, 8-bit PSD
- One normal pixel layer
- One enabled Stroke layer effect
- Position: Outside (`OutF`)
- Size: 3 px
- Opacity: 100%
- Blend mode: Normal
- Solid-color fill

```python
from psd_tools import PSDImage

psd = PSDImage.open("outside-stroke.psd")

layer = next(
    item
    for item in psd.descendants()
    if not item.is_group() and list(item.effects.find("stroke"))
)

left, top, right, bottom = layer.bbox
padding = 3
viewport = (
    left - padding,
    top - padding,
    right + padding,
    bottom + padding,
)

image = layer.composite(viewport=viewport)

print("layer bbox:", layer.bbox)
print("image size:", image.size)
print("alpha bbox:", image.getchannel("A").getbbox())

image.save("result.png")
```

For my test file, psd-tools 1.19.0 produces:

```text
layer bbox: (746, 88, 1162, 477)
image size: (422, 395)
alpha bbox: (3, 3, 419, 392)
```

The output canvas is correctly expanded from 416x389 to 422x395, but the
3-pixel padding remains transparent. The stroke is still clipped to the
original 416x389 layer bbox.

The same result occurs with psd-tools 1.12.1 and 1.19.0.

**Expected behavior**

When an expanded viewport is explicitly supplied, the outside stroke should
be rendered into that viewport.

For this test case, the alpha bounds should extend into the added 3-pixel
margin instead of remaining limited to:

```text
(3, 3, 419, 392)
```

It would also be useful if `layer.composite()` could optionally determine an
effect-aware output bbox, but at minimum an explicitly expanded viewport
should preserve the outside stroke.

**Possible cause**

The current `Compositor._apply_stroke_effect()` implementation converts the
[...]

PR #796, opened 2026-09-08
fix: draw a centered stroke effect at its full width

kyamagu:
Follow-up to #795, second half of #792.

## The defect

`draw_stroke_effect()` dilates the layer's edge with `disk(int(size / 2 - 1))`. The doubling just above it leaves an *outset* or *inset* radius a whole number, so truncation is harmless there — but a **centered** stroke keeps its half pixel, and truncating threw it away.

A 3 px centered stroke got radius 0 where it needed 1, and so reached 1 px past the layer edge where Photoshop reaches 2. Every odd size fell one pixel short per side; even sizes were already right, which is why the existing corpus never showed it — its centered strokes are all 1 px, where truncation and rounding agree.

## The fix

Round up instead. `ceil` and `int` agree whenever the operand is whole, so **no outset or inset stroke moves** — only the odd centered sizes that were wrong.

## Verification

`center-stroke-sizes.psd` is authored by scripting Photoshop and carries five squares with centered strokes of 1, 2, 3, 5 and 7 px. Photoshop's own render of that file reaches `ceil(size / 2)` px outside each square, and psd-tools now matches on every side of every square:

| size | Photoshop | before | after |
| --- | --- | --- | --- |
| 1 | 1 px | 1 px | 1 px |
| 2 | 1 px | 1 px | 1 px |
| 3 | 2 px | 1 px | **2 px** |
| 5 | 3 px | 2 px | **3 px** |
| 7 | 4 px | 3 px | **4 px** |

Testing the whole progression is deliberate rather than thorough-looking: a single 3 px case is *also* satisfied by just adding one to the radius, which would then overshoot at every even size. Five sizes distinguish the two.

Rendering the corpus in both force modes, 4 of 594 renders change — two files, both measured against their own Photoshop preview, both improving:

| Fixture | before | after |
| --- | --- | --- |
| `effects/center-stroke-sizes.psd` (new) | 0.02081 | **0.00286** |
| `effects/stroke-effects.psd` | 0.07756 | 0.07532 |

Nothing else in the corpus moves. `stroke-effects.psd` improves again but stays `xfail` — its residual error is pattern and gradient strokes. No xfail marker flips; the suite's two XPASSes (`shape-fx2`, `vector-mask2`) both predate #795.

Full suite: 2890 passed, 3 skipped, 25 xfailed, 2 xpassed. `ruff check`, `ruff format --check` and `mypy` clean.

## Still open from #792

The issue's second ask — an effect-aware output bbox, so `layer.composite()` sizes its own canvas to include effects instead of needing a hand-widened `viewport` — is an API addition and is not attempted in either PR. Worth its own issue.


kyamagu:
Both Copilot items addressed in 97f16da.

**Test helper slicing** — valid; fixed. The negative-index path was unreachable with this fixture (every square sits 16 px inside a 192×48 canvas, `_PAD` is 12), but the helper is now clamped and measures in absolute coordinates, and the failure modes clamping can hide are asserted rather than absorbed: that the window still extends the full `_PAD` past the bbox, that it contains any coverage, and that the coverage does not touch the window edge — the last two being the cases that would otherwise report an unremarked lower bound instead of a measurement. The reported reach is unchanged at every size (1, 1, 2, 3, 4 px), and the first two assertions were confirmed to fire on a near-edge bbox and an empty canvas rather than passing quietly.

**Changelog grammar** — agreed, it read as a typo in user-facing prose; now "Draw a centered stroke effect **at** its full width".

Full suite still 2890 passed, 3 skipped, 25 xfailed, 2 xpassed; ruff and mypy clean.

Issue #804, opened 2026-09-10
Stroke effects trace the viewport edge where the layer runs past it


kyamagu:
## Summary

`Compositor._apply_stroke_effect()` traces the layer from

```python
shape_in_bbox = paste(bbox, self._viewport, shape)
```

`shape` only ever covers the compositor's viewport, so wherever the layer runs
past that viewport its coverage was never read, `paste()` zero-fills it, and
the stroke measures the **viewport edge** as if it were the layer's edge. Every
stroke primitive is anchored to a boundary, so all of them inherit it: the band
(#802, #799) and the dilated edge alike.

## Reproduction 1: a layer that runs off the canvas

`tests/psd_files/effects/stroke-effect-transparent-shape.psd` is a 32x32
document holding one shape layer, bbox `(-2, -2, 34, 34)`, vector path
`(-1, -1)` to `(33, 33)`, fill opacity 0, carrying a 4 px inset stroke. The
left edge of the rectangle is one pixel off-canvas, so:

- Photoshop puts the ring at `x = -1..2`, of which `0, 1, 2` are visible.
- psd-tools sees coverage starting at `x = 0` -- the canvas edge -- and puts
  the ring at `0..3`.

Feeding `draw_stroke_effect()` the layer's own coverage placed on the stroke
bbox, with no viewport clip, and changing nothing else:

| shape fed to the stroke | MSE vs Photoshop |
| --- | --- |
| viewport-clipped (as shipped) | 0.016568 |
| the layer's own coverage | **0.0000039** |

So the whole of that fixture's remaining error is this, and nothing else.

## Reproduction 2: a viewport the caller narrows

No off-canvas geometry needed -- ask for less than the whole layer:

```python
psd = PSDImage.open("tests/psd_files/effects/center-stroke-sizes.psd")
layer = next(l for l in psd if l.name == "Size 7")   # bbox (144, 16, 160, 32)
layer.composite(viewport=(140, 12, 164, 36))   # red, blue interior, red
layer.composite(viewport=(140, 12, 152, 36))   # red, and red again at the cut
```

The second call paints a stroke down the cut, where the square simply
continues.

## Where it does and does not show

The false edge is real for every position; what differs is which side of it
the paint lands on.

- **Inset** paints inside the false edge, which is inside the viewport, so it
  is always visible -- reproduction 1.
- **Centered** paints half inside, so half of a spurious ring is visible --
  reproduction 2.
[...]

Issue #798, opened 2026-09-08
Second stroke effect on a layer traces the first stroke's outline instead of the layer


kyamagu:
## Summary

`Compositor._apply_stroke_effect` uses `shape` as both the loop variable and the accumulator, so a layer with two stroke effects draws the second stroke around the **first stroke's outline** instead of around the layer.

## Where

`src/psd_tools/composite/composite.py:1467`:

```python
for effect in _styled(layer.effects.find("stroke")):
    ...
    shape_in_bbox = paste(bbox, self._viewport, shape)
    color, shape_in_bbox = draw_stroke_effect(bbox, shape_in_bbox, effect.value, layer._psd)
    color = paste(self._viewport, bbox, color)
    shape = paste(self._viewport, bbox, shape_in_bbox)   # <-- rebinds the trace source
    opacity = effect.opacity / 100.0
    self._apply_source(color, shape, shape * opacity, effect.blend_mode)
```

The last assignment rebinds `shape` to the stroke mask just produced, so the next iteration traces that mask rather than the layer. With N strokes the error compounds.

## Measurement

Photoshop traces every stroke effect from the layer. Holding the stroke primitive constant and only changing which coverage is traced, measured against the fixture's own Photoshop preview (MSE):

| tracing | MSE vs Photoshop |
| --- | --- |
| chained (current) | 0.05653 |
| each stroke traced from the layer | **0.00227** |

A 25x error reduction, which is essentially the entire error on that fixture.

`tests/psd_files/effects/double-stroke-effects.psd` is the affected fixture — one shape layer carrying an `OutF` and an `InsF` stroke. It is the only layer in the whole fixture corpus with more than one stroke effect, which is why this has gone unnoticed.

## Consequence for the test suite

That fixture is currently listed in `test_stroke_effects_xfail` (`tests/psd_tools/composite/test_effects.py`) with `threshold=0.01`. At 0.00227 it passes, so this fix retires an `xfail` rather than merely improving a number.

## Suggested fix

Keep the traced coverage in a name distinct from the per-effect result, so each iteration traces the layer:

```python
traced = shape
for effect in _styled(layer.effects.find("stroke")):
    ...
    shape_in_bbox = paste(bbox, self._viewport, traced)
    color, mask = draw_stroke_effect(bbox, shape_in_bbox, effect.value, layer._psd)
    color = paste(self._viewport, bbox, color)
    mask = paste(self._viewport, bbox, mask)
[...]

PR #817, opened 2026-09-16
fix: trace a stroke effect on a group from the group itself (#808)

### LINKED ISSUE #808: composite: strokes at a group boundary still trace the wrong box (#804 remainder)
## Summary

#804/#806 fixed a stroke effect tracing the compositor's viewport edge instead
of the layer's boundary, by reading the layer's coverage again on the box the
stroke draws on. Two cases in the same family are still open, both involving a
group, and both left alone deliberately in #806.

## 1. A stroke effect *on* a group still traces the viewport

`_trace_shape()` re-reads a layer's coverage on the stroke's box. A group has
no coverage to re-read: a group's coverage *is* the composite onto the
compositor's viewport and exists nowhere else. Reading it as an object finds no
pixel data and no fill, which would place the stroke on an empty canvas and
draw nothing at all — worse than a misplaced stroke — so #806 carved groups out
and left them on the clipped copy:

```python
if isinstance(layer, GroupMixin) and not traces_mask:
    return paste(viewport, self._viewport, shape)
```

So a group whose bbox runs past the compositor's viewport still gets its stroke
anchored to the viewport edge, exactly as #804 describes.

Doing this properly means compositing the group a second time on the wider box,
which is expensive and recurses through nested groups — worth designing rather
than bolting on. No fixture in the corpus puts a stroke effect on a group, so
nothing measures this today; a fixture is the first thing needed.

## 2. A non-passthrough group's viewport cuts a child's outset stroke

`_get_group()` gives a non-passthrough group

```python
viewport = utils.intersect(self._viewport, layer.bbox)
```

and `Group.bbox` is the union of its children's bboxes, which excludes effect
coverage — the same box #797 is about, one level up. A child's outset or
centered stroke reaches past its own bbox and therefore past the group's, and
is clipped away during an ordinary document composite, with no viewport the
caller could widen to get it back.

The mechanism, using the fixture #792 was fixed on — a 16×16 square with a 3 px
outset stroke, where `layer.bbox` is `(8, 8, 24, 24)` and the stroke's box is
`(4, 4, 28, 28)`:

```python
tight = composite(layer, viewport=layer.bbox, as_layer=True)[2]   # what a tight group hands it
wide  = composite(layer, viewport=(4, 4, 28, 28), as_layer=True)[2]
tight.sum()   # 256.0  -- the square, and none of the stroke
wide.sum()    # 473.86
```

The whole stroke, 218 of 474 alpha units, is outside the box a tight group
would give it.

I could not find a corpus fixture where this loses visible paint. The nearest
is `advanced-blending.psd`: group `Frame`, bbox `(668, 34, 1166, 591)`,
non-passthrough, whose child `Rectangle 1` has a stroke box of
`(668, 34, 1167, 592)` — one pixel over on the right and bottom. Widening
`Frame` by that pixel changes the render by exactly nothing, because the pixel
in question is the fixed margin `stroke_bbox()` adds for the layer's *edge*
rather than for the stroke, and carries no paint. So this is a real structural
hole with no fixture that exhibits it, and authoring one is part of the work.

Related: #797 asks for an effect-aware output bbox on `layer.composite()`.
That is the caller-facing half of the same arithmetic; this one is internal and
a caller cannot work around it.

Found while reviewing #806.

kyamagu:
Closes #808. Item 2 shipped as #816; this is item 1, the remainder.

## The defect

#804 made a stroke effect re-read its layer's own coverage on the box it draws
on, so it traces the layer's boundary rather than the compositor's viewport
edge. Groups were carved out and left tracing the clipped copy: a group has no
stored pixels and no fill, so reading it as an object finds no coverage at all
and would draw nothing — worse than a misplaced stroke.

## The fix

`_get_group_shape()` composites the group's contents again onto the box asked
for. It narrows that box exactly the way `_get_group()` narrows it, and the
asymmetry is load-bearing in both directions:

- an **isolated** group is intersected with `_content_bbox()`, which returns an
  `Artboard`'s frame verbatim — that intersection *is* the artboard clip (#816),
  so skipping it would trace the children out past the frame;
- a **pass-through** group gets the box untouched, because `_get_group()` gives
  one this compositor's viewport untouched. A pass-through artboard therefore
  gets no frame clip today — `artboard.psd`'s children reach `(-736, -356)`, and
  16k pixels of its paint live outside its frame. The trace reproduces that
  rather than improving on it.

Both directions are asserted.

## The fixture

No fixture in the corpus puts a stroke effect on a group (swept: 0 of 298), so
`group-stroke-off-canvas.psd` is authored in Photoshop 2026 and its own preview
is the ground truth. 96x32, two groups, each with a 4 px inset stroke *on the
group*.

`Clipped` is an L of two children shifted 12 px off the left edge, so only the
upper arm crosses `x = 0` and the boundary through the escaping strip is
horizontal. That geometry is the whole point — it separates three answers at two
rows:

| | row 8 (through the arm) | row 20 (below it) |
|---|---|---|
| Photoshop | no band — the group continues off-canvas, so there is no edge | band at `x = 0..3` |
| before | **band** — traced its own clipped copy's false edge | band |
| "treat outside the viewport as covered" | no band | **no band** — erases a real edge |

`Inside` is the control: its stroke box stays within the canvas, so it takes the
early return and is asserted byte-identical to Photoshop's render.

Document MSE **0.012712 → 5.5e-06**, identical in both force modes.

Two earlier fixtures did not survive their own mutation testing and are worth
recording: a group whose child filled its bounding box exactly was passed by two
implementations that do **no compositing at all**, and a disc failed too — a
circle is at its widest where it crosses the canvas edge, so its silhouette
there reads exactly like a rectangle's.

## Cost

Tracing composites a subtree, and a group inside whose own stroke reaches
further than its parent's composites its subtree again, so the work can double
per level. Equal stroke sizes all the way down do *not* do it — those inner
traces stay inside the box #816 already widened — but with sizes increasing
inward a chain of ten costs 1024 composites. Memoized per pass on
`(group, box)`, keyed by box because a layer may carry several stroke effects
(#798). Both memos now travel as one `_PassCache`.

Peak allocation, measured with a forged stroke size of 1500: **unchanged** for an
isolated group, because its inner box follows the contents rather than the
descriptor's margin; **2.0x** for a pass-through group, which has no such bound.
The memo is also a new retention — one coverage canvas per entry until the pass
ends — and that is written down in its docstring rather than left implicit.

## Verification

- All **4240** corpus renders — every fixture, per document *and* per layer, in
  both force modes — byte-identical. No shipped fixture reaches the new branch,
  so the new fixture and the unit tests are the whole of its coverage.
- Five targeted reverts each fail, each caught by the test written for it: the
  pre-fix clipped copy, forgetting `* _get_mask`, dropping the frame clip,
  clipping a pass-through group, and the memo never hitting.
- `uv run pytest` 2966 passed; `ruff check`, `ruff format --check`, `mypy` clean.

## Scope notes

- A group whose stroke traces its **mask** (`force` plus a vector mask on the
  group) still gets the mask outline alone, which for a group bears no relation
  to where its contents are. Pre-existing, unreachable in the corpus, and the
  same inside the viewport as outside, so it is left alone with a TODO naming it.
- The changelog also drops "A stroke on a group is unchanged" from the #804/#806
  entry, which would otherwise ship in the same release as the commit that makes
  it false.


kyamagu:
Taken — fixed in d0d70df8. The suppressed comment has no inline thread to reply
into, so it is answered here.

"an artboard still clips to its frame" was wrong in exactly the way described: a
*pass-through* artboard has no frame clip today, because `_get_group()` hands a
pass-through group this compositor's viewport untouched, and
`_get_group_shape()` reproduces that on purpose rather than improving on it —
`test_a_traced_artboard_keeps_its_frame_clip` asserts both directions, including
the 16k pixels of `artboard.psd`'s paint that live outside its frame and do
render.

Rather than qualify it to isolated artboards, the entry now says the trace clips
exactly as the composite does, artboards included. That is shorter, true of every
group, and stays true if the pass-through artboard quirk is ever fixed on its
own.

kyamagu:
Both suppressed comments taken, in 76ed2fcd.

**`test_widen.py`'s propagation invariant.** Right, and its own docstring says a
site that stops being reached should fail there rather than go quiet — so a site
that was never reached is the same hole. `_get_group_shape` is added to
`_COMPOSITOR_CALLERS`, and `effects/group-stroke-off-canvas.psd` to the fixture
list, because no existing fixture puts a stroke effect on a group and so none
reached it. Verified by dropping `widen=` and `color_mode=` from the new site:

```
FAILED test_every_sub_compositor_inherits_the_document_facts
```

**The memo's box key.** Also right, and the first test I wrote for it proved
nothing — worth recording. It asked the *isolated* group for two different
boxes, but both narrow to the same `_content_bbox()`, so they legitimately share
one cached array and only the `paste()` back differs. Keyed on `id(layer)` alone
it still passed.

A pass-through group is where the key actually bites: `_get_group_shape()` hands
it the box untouched, so two boxes really do composite on different canvases and
a shared entry would be the wrong size. `test_a_traced_group_keeps_its_boxes_apart`
now uses `masks3.psd`'s pass-through group and asserts two cache entries, and
the same mutant fails:

```
FAILED test_a_traced_group_keeps_its_boxes_apart
```

Suite 2967 passed; `ruff`, `ruff format`, `mypy` clean.

Issue #814, opened 2026-09-16
Adding layers to a group does not invalidate its cached bbox


kyamagu:
## Summary

`GroupMixin.append()`, `extend()` and `insert()` never invalidate the group's
cached `_bbox`, so a group populated through the editing API reports
`(0, 0, 0, 0)` for the rest of its life — and so do `width`, `height`, `left`,
`top`, `right` and `bottom`.

A group is born with the empty box already cached, so there is no ordering that
avoids it. `psd_image.create_group(layer_list=...)` is affected too, since it
extends the group after creating it.

## Reproduction

psd-tools 1.19.0, on a fixture in this repo:

```python
from psd_tools import PSDImage
from psd_tools.api.layers import Group

for how in ("create_group(layer_list)", "create_group() + append", "create_group() + insert"):
    psd = PSDImage.open("tests/psd_files/effects/outside-stroke.psd")
    layer = psd[0]
    if how == "create_group(layer_list)":
        group = psd.create_group([layer], name="G")
    elif how == "create_group() + append":
        group = psd.create_group(name="G"); group.append(layer)
    else:
        group = psd.create_group(name="G"); group.insert(0, layer)
    print(how, len(group), group.bbox, Group.extract_bbox(group))
```

```
create_group(layer_list)     1 (0, 0, 0, 0) (8, 8, 24, 24)
create_group() + append      1 (0, 0, 0, 0) (8, 8, 24, 24)
create_group() + insert      1 (0, 0, 0, 0) (8, 8, 24, 24)
```

`Group.extract_bbox()` computes the right answer every time; only the cached
property is stale.

## Where it comes from

`GroupMixin.bbox` memoizes into `self._bbox` and the three mutators do not call
`Layer._invalidate_bbox()`. `__setitem__`, `__delitem__`, `clear()` and
`remove()` are worth auditing in the same pass.

The cache is seeded before any layer is added: `_update_children()` does
`isinstance(layer, GroupMixin)`, and `GroupMixin` is a `@runtime_checkable`
`Protocol`, so that check reads `bbox` and materializes `(0, 0, 0, 0)`. A freshly
created `Group` therefore already has `_bbox` set:

```python
g = Group.new(parent=psd, name="Fresh")
g._bbox   # (0, 0, 0, 0), before anything is appended
```

Setting `visible` or `clipping` *does* invalidate correctly
(`Layer._invalidate_bbox()` walks to the root), so the machinery exists and the
mutators simply do not use it.

## Impact

[...]

PR #821, opened 2026-09-16
fix: invalidate descendants' cached bboxes when a group is hidden

### LINKED ISSUE #819: Hiding a group leaves its descendants' cached bboxes stale
## Summary

`Group.extract_bbox()` filters children on `layer.is_visible()`, which walks
**up** the parent chain — so a group's bounding box is a function of its
*ancestors'* `visible` flags. But `_invalidate_bbox()` only walks up too, so
hiding a group never drops the boxes cached on its descendants.

The result is a child reporting a box strictly larger than its parent's, and an
answer that depends on the order the boxes were first read.

## Reproduction

On `tests/psd_files/clipping-mask.psd`, at f437798d and on #818 alike:

```python
from psd_tools.api.layers import Group

psd = PSDImage.open("tests/psd_files/clipping-mask.psd")
g2 = psd[1]        # 'Group 2'
g1 = g2[0]         # 'Group 1', nested inside it

armed = g1.bbox    # (103, -73, 288, 146)
g2.visible = False

print(g1.bbox)                    # (103, -73, 288, 146)  <- stale
print(Group.extract_bbox(g1))     # (0, 0, 0, 0)          <- truth
print(g2.bbox)                    # (0, 0, 0, 0)
```

The child now claims a box its hidden parent does not. Read order decides the
answer — on a fresh document where nothing was cached first:

```python
psd = PSDImage.open("tests/psd_files/clipping-mask.psd")
psd[1].visible = False
print(psd[1][0].bbox)             # (0, 0, 0, 0)
```

## Where it comes from

`Layer._invalidate_bbox()` clears `self._bbox` and recurses into `self.parent`.
That is the right direction for a *contents* change, which only affects
ancestors. Visibility is the other direction: `is_visible()` is
`self.visible and (parent is None or parent.is_visible())`, so toggling a
group's flag changes the extracted box of everything beneath it.

The same applies to any ancestor-dependent filter. `clipping` is not affected —
it is read on the child itself.

## Impact

Lower than #814: it needs a group's box to have been read before an ancestor is
hidden, and `composite()` re-reads through the visibility check anyway. But a
caller that reads `group.bbox`, `width`, `height`, `left`, `top`, `right` or
`bottom` after hiding an ancestor gets a box for content that will not render.

## Notes

Pre-existing and **not** a regression from #818, which fixes the upward half of
the same cache. Filed out of that work so the downward half is not mistaken for
something #818 introduced.

A fix would drop descendants' boxes from the `visible` setter — bounded by the
subtree, and only when the flag actually changes. Worth checking against the
recursion-depth note in #818 before making it recursive.

kyamagu:
Closes #819.

`Group.extract_bbox()` filters children through `is_visible()`, which walks
**up** the parent chain — so a container's bounding box is a function of its
*ancestors'* `visible` flags. `_invalidate_bbox()` only walks up too, so hiding
a group never dropped the boxes cached beneath it. A nested group kept
reporting a box strictly larger than its hidden parent's, and the answer
depended on the order the boxes were first read:

```python
psd = PSDImage.open("tests/psd_files/clipping-mask.psd")
g2, g1 = psd[1], psd[1][0]

armed = g1.bbox    # (103, -73, 288, 146)
g2.visible = False

g1.bbox                    # (103, -73, 288, 146)  <- stale, before
Group.extract_bbox(g1)     # (0, 0, 0, 0)          <- truth
g2.bbox                    # (0, 0, 0, 0)
```

This is the downward half of the cache #818 fixed the upward half of. It is
pre-existing and not a regression from that PR.

`Group` overrides `left`, `top`, `right` and `bottom` to read `self.bbox`, and
`width`/`height` derive from those, so the stale value surfaces through seven
public properties. It also reaches **rendered output**: `layer.composite()`
takes the layer's own `bbox` as its viewport, so the descendant was composited
onto the box it had while visible.

```
                          armed first          read fresh
inner.composite(force=True)   185x219  08e7e7a2    360x200  c363cfe8   DIFFER
psd.composite()               360x200  fb5989c9    360x200  fb5989c9   agree
```

`psd.composite()` never differed — it re-derives through the visibility filter
rather than reading a descendant's cache, and nothing consults a descendant's
cache on the way down (`_get_bbox()` recurses through `Group.extract_bbox()`,
not through `child.bbox`), so a parent's box stays correct too.

An earlier revision of this PR claimed the staleness was "observable only on a
direct read of the descendant". That was wrong, and review caught it; the
measurement above is the correction, and it narrows the scope as well as
widening it.

## The change

All three parts are in `Layer.visible`'s setter.

**The fix** is a downward walk, reusing `_invalidate_subtree_bbox()` — the
helper #818 added for reparenting — rather than growing a third near-identical
walk. `Group` is named rather than `GroupMixin` because the latter is a
`runtime_checkable` protocol whose `isinstance` runs `hasattr(x, "bbox")` on
Python ≤3.11, recomputing the subtree a line before the walk drops it. Review
was right that this is work avoided, not correctness — measured, the two
guards give identical answers and `GroupMixin` costs one extra
`extract_bbox()`. `Artboard` is a `Group` subclass, so the walk passes
through it either way.

It over-clears slightly: `_invalidate_subtree_bbox()` also drops
`ShapeLayer._bbox`, which tracks the *document's* size rather than visibility.
Deliberate — a dropped cache is never wrong, the recompute returns an equal
value, and a hidden subtree is not read by `extract_bbox()` or `composite()`
anyway, so those caches mostly never recompute. A fourth walk that visited only
containers would buy a near-zero saving for another near-identical name.

**An early return** when the flag does not change, which is what keeps the new
walk bounded. Nothing in the public API writes `_record.flags.visible` except
this setter, and nothing in `src/` assigns `.visible` outside two module
docstring examples, so no cache can go stale behind that guard. The exception
is writing `_record.flags.visible` directly and then assigning the same value
through the setter — that used to get an incidental upward invalidation and
now gets none. It is the documented out-of-contract case (stale high-level
caches after low-level mutation), and no test depends on it. It also drops a
pre-existing wart: `visible = "x"` used to mark an already-visible document
dirty, because `True != "x"`.

**Writing the flag before the walks** rather than after. This one is ordering
hygiene, **not** a fix — both walks discriminate on concrete classes, so
neither re-reads a box today, and reverting just this part leaves every test
green (checked). Called out because it is behaviour-neutral and no test pins
it.

## Recursion depth — the open question on the issue

The issue asked whether a downward walk would make the ≤3.11 recursion ceiling
worse, since invalidating more means the next read actually recomputes.
Measured on 3.10 over nested groups, the ceiling is **unchanged**:

| nesting depth | without downward walk | with downward walk |
| ------------- | --------------------- | ------------------ |
| 80            | ok                    | ok                 |
| 90            | `RecursionError`      | `RecursionError`   |

and the traceback bottoms out in `is_visible()` — the existing *read* path, not
the walk, which costs ~2 frames per level against a read path that already
costs far more. So no mitigation was needed.

## Tests

Seven, all read-then-mutate (arming the cache before hiding), in a documented
section alongside the #814 ones. Each was mutation-tested, and four of the five
mutants are killed by exactly one test:

| mutant                                   | killed by                                            |
| ---------------------------------------- | ---------------------------------------------------- |
| no downward walk at all                  | 6 of the 7                                           |
| walk one level, no recursion             | `..._reaches_every_level_beneath_not_just_the_first` |
| invalidate only when hiding, not showing | `..._showing_a_group_again_restores_...`             |
| `type(self) is Group` (misses `Artboard`) | `..._hiding_an_artboard_drops_its_descendants...`    |
| no early return                          | `..._setting_visible_to_the_value_it_already_has...` |
| old ordering (flag written last)         | nothing — see above                                  |

All seven fail against unfixed source on **both** 3.10.21 and 3.14.7.

The three-level test uses `adjustment_nested_composition_4.psd`, which ships
three nested groups that all have a non-empty box, rather than building the
tree with `create_group()` — that setup would have exercised #818's reparenting
invalidation, which drops the same subtree, leaving a reader unable to tell
which mechanism the test measures. The render test is marked `composite` so the
no-composite CI builds deselect it (verified: 1 deselected).

## Verification

- 2993 passed, 3 skipped, 19 xfailed, 2 xpassed on 3.10.21 **and** 3.14.7
- Cost of the walk, since hiding every group in turn makes it O(subtree) each:
  synthetic trees of 20/84/258/340 groups give a flat **~2× on the setter**
  (0.36 ms against 0.17 ms at 340 groups), not a degrading ratio.
- `ruff check`, `ruff format --check`, `mypy` clean
- Corpus sweep over the 86 fixtures that contain groups: arm every cached box,
  hide one group, then diff every cache against `Group.extract_bbox()`. Over
  9377 checks, **178 stale before the fix, 0 after**.
- The same sweep toggling `clipping` instead — read on the child, not on an
  ancestor — gives **0 stale over 18676 checks**, confirming `visible` is the
  only ancestor-dependent input that needed a downward walk.
- `Artboard` is excluded from that oracle, not skipped by the fix: its `bbox`
  is the `artboardRect` from its tagged blocks rather than a union over
  children, so `extract_bbox()` is not its truth. The walk still passes
  *through* it, which is what the artboard test pins.


kyamagu:
Checked the "avoidable full-subtree traversal on every show" point. It doesn't hold — the quoted line can't reach that `isinstance`.

`isinstance(self, (GroupMixin, ShapeLayer))` at `layers.py:211` lives in **`Layer._invalidate_bbox`**. A `Group` never reaches it, because `GroupMixin` precedes `Layer` in the MRO:

```
Group._invalidate_bbox -> GroupMixin
GroupMixin._invalidate_bbox contains isinstance: False
Layer._invalidate_bbox  contains isinstance: True
```

`GroupMixin._invalidate_bbox` is `self._bbox = None` plus a walk to the parent, and every container above it resolves the same way, so no protocol check runs anywhere on that upward chain.

Measured on **3.10.21** — the interpreter where the probe is real — counting `Group.extract_bbox` invocations across the whole setter on a synthetic 84-group tree:

```
hide (True -> False): extract_bbox calls=0    4.4 us
show (False -> True): extract_bbox calls=0    3.3 us
```

Zero in both directions, so the ordering of the flag write doesn't introduce a traversal either. For completeness, the case that *does* reach `Layer._invalidate_bbox` — a non-group layer — also measures 0 `bbox` executions and 0 `extract_bbox` calls, and has no subtree to traverse regardless.

Leaving the upward helper as it is. Narrowing its `isinstance` to concrete classes would be a pre-existing micro-optimisation unrelated to this fix, and #818 showed that changing those checks needs its own measurement rather than being folded into an unrelated PR.

The mechanism you're pointing at is real on ≤3.11 — `isinstance` against a `runtime_checkable` protocol does run `hasattr` and execute the `bbox` descriptor, which is exactly why this PR's downward guard names `Group` concretely. It just isn't reachable from the line quoted.

Issue #826, opened 2026-09-17
composite: a malformed stroke-effect descriptor is tolerated when measured but raises when drawn


kyamagu:
## Summary

`_stroke_reach()` guards the descriptor reads it makes and degrades to
`layer.bbox` when they fail, so a malformed stroke effect cannot break the box
a group composites on or the viewport cull. The code that goes on to *draw*
that stroke makes the same reads unguarded, so the same descriptor still takes
down the composite.

The asymmetry is the bug: measuring a stroke effect is total, drawing it is
not. `_stroke_reach()`'s own comment says a bad descriptor "must not turn a
document that rendered into one that raises", and for the measurement it does
not — but `Compositor._apply_stroke_effect()` calls `stroke_bbox()` directly
at `composite.py:1837`, and `draw_stroke_effect()` reaches
`paint._get_color()`, neither of which is guarded.

## Reproduction

psd-tools 1.19.0. Each row mutates the one stroke effect on
`effects/outside-stroke.psd` and then calls `psd.composite()`:

| defect | `_stroke_reach()` | `psd.composite()` |
|---|---|---|
| `del desc[Key.Style]` | degrades to `bbox` | `AttributeError` — `effects.py:109` |
| `desc[Key.SizeKey] = "wide"` | degrades to `bbox` | `ValueError` — `effects.py:114` |
| `desc[Key.Style] = Double(1.0)` | degrades to `bbox` | `AttributeError` — `effects.py:109` |
| `del desc[Key.Color]` | unaffected | `ValueError` — `paint.py:268` |
| `desc[Key.Color].classID = b"XXXX"` | unaffected | `KeyError` — `paint.py:269` |
| `del desc[Key.SizeKey]` | fine | renders |
| `del desc[Key.Opacity]` | fine | renders |

```python
from psd_tools import PSDImage
from psd_tools.terminology import Key

psd = PSDImage.open("tests/psd_files/effects/outside-stroke.psd")
layer = psd[0]
for effect in layer.effects.find("stroke"):
    del effect.value[Key.Style]

psd.composite(ignore_preview=True)   # AttributeError: 'NoneType' object has no attribute 'enum'
```

Three things worth separating out:

- **`Key.Style` has no default on the very line where `Key.SizeKey` has one.**
  `stroke_bbox()` reads `desc.get(Key.Style).enum` but
  `desc.get(Key.SizeKey, 1.0)`, so a missing size is tolerated and a missing
  style is not, two tokens apart.
[...]

Issue #786, opened 2026-09-02
chore: adopt the remaining pydocstyle rules (D1, D200, D401)


kyamagu:
Follow-up to #772, which enabled ruff's `D` family with three rules held back. Splitting them out so the `ignore` entries in `pyproject.toml` have a live tracking pointer rather than referring back to the closed issue that deferred them.

Per the #772 thread, all three are excluded on noise grounds rather than principle and are "longer-term scope". Counts measured repo-wide with ruff 0.16.4 and `convention = "pep257"` — note these are larger than the figures in #772, which covered `src/` and `tests/` only while `ignore` applies to the whole tree.

| rule | violations | notes |
| --- | --- | --- |
| `D1` missing docstrings | **1121** | 544 `D102` undocumented-public-method, 373 `D103`, 95 `D105`, 42 `D100`, 40 `D101`, 21 `D107`, 6 `D104` |
| `D401` non-imperative-mood | **96** | no fixes available |
| `D200` unnecessary-multiline-docstring | **87** | no *safe* fixes; all 87 behind `--unsafe-fixes` |

Suggested order, smallest and least contentious first:

1. **`D200`** — mechanical, but every fix is unsafe-only, so each wants reading. Now that #770 has compressed the docstring prose, the churn objection from #772 no longer applies.
2. **`D401`** — the real question is whether to take it at all. It misfires on the property docstrings this codebase has many of ("Gradient type, one of ...", "Kind of this layer."), where the imperative mood is wrong rather than merely unidiomatic. A per-file `per-file-ignores` for the property-heavy modules may be the better shape than a tree-wide adoption.
3. **`D1`** — much the largest, and a different project from formatting the docstrings that exist: it is the rule most readily satisfied with `"""Return the value."""`. Worth deciding whether the target is all of `D1` or only `D100`/`D101`/`D104` (88 violations between them, the module- and class-level ones), leaving `D102`/`D103` off.

No urgency on any of these; the tree is green with them ignored.

Issue #772, opened 2026-08-31
proposal: enable ruff's docstring rules (pydocstyle D) as a scoped subset


kyamagu:
Proposal to enable ruff's `D` (pydocstyle) rules, as a scoped subset. Splits out of
#692, which evaluates the whole expanded default rule set; this covers only the
docstring family, which came up while measuring docstring prose for #770 and #771.
Measured with ruff 0.16.4 against `main` at 8d44ed0.

## What this is not

It is not a verbosity control, which is what prompted the question. **No `D` rule
measures docstring length.** The closest by name, `D200
unnecessary-multiline-docstring`, points the other way -- it collapses a *short*
docstring written across several lines. So none of this would have flagged
`_image_data_planes()` at 71 prose lines over 13 of code, and it does not replace
the release-time sweep in #771. `D` governs the shape of a docstring; the sweep
governs how much of it there is.

## Proposal

```toml
[tool.ruff.lint]
select = ["E4", "E7", "E9", "F"]
extend-select = ["PLC0415", "T10", "D"]
ignore = [
    "D1",    # missing docstrings -- a separate project, see below
    "D200",  # collapse short multi-line docstrings -- churn, no fixes available
    "D401",  # imperative mood -- opinionated, and wrong for property docstrings
]

[tool.ruff.lint.pydocstyle]
convention = "pep257"
```

`pep257` is the load-bearing choice, on two grounds. It matches the codebase's
reST/Sphinx docstrings (`:param x:`), which neither the google nor the numpy
convention describes. And it disables `D212 multi-line-summary-first-line`, which
under `google` alone is **452 violations** -- a reflow of nearly every multi-line
docstring in the tree, for a stylistic preference.

## What it would flag

104 violations across `src/` and `tests/`, 23 of them safely auto-fixable and 22
more with `--unsafe-fixes`:

| rule | count | |
| --- | --- | --- |
| `D205` missing-blank-line-after-summary | 55 | manual |
| `D400` missing-trailing-period | 16 | manual |
| `D202` blank-line-after-function | 14 | auto |
| `D301` escape-sequence-in-docstring | 7 | manual |
| `D209` new-line-after-last-paragraph | 5 | auto |
| `D403` first-word-uncapitalized | 3 | auto |
| `D402` signature-in-docstring | 2 | manual |
| `D412` blank-lines-between-header-and-content | 1 | auto |
| `D419` empty-docstring | 1 | manual |

[...]

Issue #692, opened 2026-08-02
Evaluate adopting ruff 0.16's expanded default rule set


kyamagu:
## Background

Ruff 0.16.0 ([blog post](https://astral.sh/blog/ruff-v0.16.0)) shipped two breaking changes to its defaults:

1. **The default lint rule set expanded from 59 to 413 rules.**
2. **The formatter now formats Python code blocks inside Markdown by default.**

Both surfaced on the ruff dependency bump in #690. To keep that bump behaviour-neutral, #690 pinned the lint selection to the previous defaults and excluded `*.md` from the formatter:

```toml
[tool.ruff.format]
exclude = ["*.md"]

[tool.ruff.lint]
select = ["E4", "E7", "E9", "F"]
extend-select = ["PLC0415", "T10"]
```

This issue tracks deliberately evaluating and (selectively) adopting the new default rules, which is a much larger change that shouldn't ride along with a dependency bump.

## Scope

Running `ruff check --statistics` against the expanded default set reports **526 findings** (288 auto-fixable with `--fix`, plus 103 hidden fixes behind `--unsafe-fixes`). The largest categories:

| Count | Rule | Description | Auto-fix |
| ----: | ---- | ----------- | :------: |
| 125 | UP031 | printf-string-formatting (`%` → f-strings) | manual |
| 96 | PYI019 | custom-type-var-for-self | ✓ |
| 45 | UP035 | deprecated-import | ✓ (unsafe) |
| 41 | I001 | unsorted-imports | ✓ |
| 30 | UP006 | non-pep585-annotation (`List` → `list`) | ✓ |
| 21 | PIE796 | non-unique-enums | manual |
| 19 | UP037 | quoted-annotation | ✓ |
| 17 | PIE790 | unnecessary-placeholder | ✓ |
| 15 | C408 | unnecessary-collection-call | manual |
| 11 | UP008 | super-call-with-parameters | ✓ |

...plus a long tail of `UP*`, `SIM*`, `RUF*`, `PYI*`, `B*`, `S*` rules in single digits.

## Suggested approach

- Enable the rule families incrementally (e.g. `I` import sorting, then `UP` pyupgrade, then `SIM`/`PIE`/`RUF`) so each change is reviewable in isolation.
- Apply `ruff check --fix` per family and manually review the remainder.
[...]

PR #829, opened 2026-09-17
fix: skip an effect psd-tools cannot read, and stop a repr raising

### LINKED ISSUE #828: composite: an unknown effect class makes Layer.__repr__ raise, which aborts every composite
## Summary

`Layer.__repr__` calls `has_effects()`, which builds `Effects`, which raises
`ValueError` on an effect class it does not know
(`api/effects.py:71`). So `repr(layer)` raises on such a file — and because
`Compositor.apply()` formats every layer it visits eagerly at
`composite.py:1020`:

```python
logger.debug("Compositing %s" % layer)
```

the whole document is unrenderable, at any log level, including with logging
disabled outright.

## Reproduction

psd-tools 1.19.1.dev (main at #827). The mutation forges an effect class the
`_TYPES` registry does not hold; `del layer._effects` is only there because
`Layer.effects` memoises.

```python
import logging
logging.disable(logging.CRITICAL)

psd = PSDImage.open("tests/psd_files/effects/outside-stroke.psd")
layer = psd[0]
layer.effects._data[b"FrFX"].classID = b"XXXX"
del layer._effects

repr(layer)            # ValueError: Effect class not found for b'XXXX'
layer.has_effects()    # ValueError
psd.composite()        # ValueError, from composite.py:1020
```

Substituting a `__repr__` that cannot raise is enough to make that same
document composite:

```python
import psd_tools.api.layers as L
L.Layer.__repr__ = lambda self: "<layer>"
psd.composite(ignore_preview=True)   # renders
```

## Two separable defects

- **A repr that raises.** `__repr__` is called by debuggers, by `print()`, by
  every `%`-formatted log line in the compositor and by pytest when an
  assertion fails. Whatever `has_effects()` should do, `repr(layer)` reporting
  `" effects"` or not is never worth an exception.
- **Eager formatting on a hot path.** `composite.py:1020` builds the string
  whether or not DEBUG is enabled, which is both the cost and the reason the
  raise is unconditional. `_apply_overlay()` at `composite.py:1699` already
  passes its arguments lazily; most of the module does not.

There is also a third question behind them — whether `Effects.__init__`
should reject an unknown effect class at all, or skip it the way it already
skips one that is not present (`api/effects.py:65-66`). Rejecting is defensible
for a constructor; raising out of a *repr* is not.

## Not reachable from a Photoshop file

Every effect class in `tests/psd_files` is one `_TYPES` holds, so no shipped
fixture reproduces this and the mutation above is forged. Robustness for files
psd-tools did not write, not a reported user failure.

## Why it is filed apart from #826

#826 was about an effect whose *descriptor* cannot be read, and #827 fixed
that: `_readable()` (`composite.py:67`) already degrades to no effects for this
very `ValueError`, and the drawing loops guard each effect they read. What
remains is not an effect that cannot be drawn but a repr that raises, one
frame above and in a different subpackage. #827 left ~17 lines of working
guard that no composite can reach because of it.

Related: #826, #827.

kyamagu:
Closes #828.

`Layer.__repr__` calls `has_effects()`, which builds `Effects`, which rejected
an effect class it had no handler for — and `Compositor.apply()` formatted
every layer it visited with `%` before handing it to `logger.debug`. So one
effect class psd-tools does not know made a whole document unrenderable, at
any log level, logging disabled outright included.

Four parts, in the order the failure travels.

## `Effects` skips what it cannot interpret

`Effects.__init__` now skips such an effect, the way it already skips one that
is not present five lines above. Rejecting it is defensible for a constructor
read on its own, which is why #828 filed it as a question rather than a bug —
but `Effects` is built on first access from read-only paths that can only pass
a raise on. One effect lost, rather than the document.

`_readable()`, which #827 added one frame below for this very `ValueError`,
keeps its guard and loses the cause it was written for. Its docstring says so
now; the clause stays because the listing is what the guards downstream stand
on, and being total about it costs one `try`.

## A second way into the same constructor, found in review

`TaggedBlock.read()` keeps the raw bytes of a block it could not parse, so
`self._data` could be `bytes`, and reading those as a descriptor raised. This
one needs no forged class name — only an effects descriptor that will not
read, which the low-level reader already degrades to raw bytes and logs about,
so it is the more reachable of the two. Measured on `main`:

| forged effects block | before |
|---|---|
| `b"garbagebytes"` | `IndexError` — `api/effects.py:62` |
| `b"\x00" * 300` | `AttributeError` — `api/effects.py:88`, via `enabled` |

Which exception it was depended on the bytes, and that is the point:
`IndexError` is not in `_UNREADABLE`, so for that content no guard #826 added
could see it. Both are now the no-effects case, which every property already
answers for — `scale` by raising, exactly as it already does for a layer
carrying no effects block at all.

## A repr that reports what it can read

`Layer.__repr__` drops an annotation it cannot read rather than raising. Two
of its parts read beyond the record's own fields, and both could raise for a
reason of their own:

- `has_effects()` parses the effects descriptor — the defect above.
- `Artboard.bbox`, which `self.width` reaches, raises outright when the
  artboard's tagged block is absent. **This is a second instance of #828's
  first defect that the issue does not name**, one class away and nothing to
  do with effects:

  ```
  tests/psd_files/artboard.psd, ARTBOARD_DATA1 deleted
  repr raised: ValueError Artboard data not found in tagged blocks
  ```

`except Exception`, not `_UNREADABLE`: that tuple is narrow because its caller
goes on to *draw*, so an exception outside the set is a bug worth surfacing
rather than wrong pixels worth hiding. A repr draws nothing, and nothing that
calls one — `print()`, pdb, a logging handler, pytest — asked what it reads.
`test_repr_drops_an_annotation_it_cannot_read` is parametrized over one
exception the tuple holds and one it does not, so narrowing the clause fails
exactly the case that argues for the wide one.

`name`, `visible`, `clipping` and `has_mask()` stay unguarded, because they
read already-parsed record fields. `PSDImage._repr_pretty_` — what IPython
calls to display a whole document, and which raised for the same reason —
inherits the fix for free; `PSDImage.__repr__` reads only the header and needs
nothing.

Whether such an artboard's *document* composites is deliberately not
asserted, and finding out why cost a CI round: it raises on Python <= 3.11 and
renders on 3.12. The `isinstance(layer, GroupMixin)` in `_resolve_source()`
runs `hasattr(x, "bbox")` on the older protocol implementation, and `hasattr`
swallows only `AttributeError` — so the artboard's `ValueError` escapes from a
type check there and nowhere else. That is the footgun `_accepts()` already
documents, and it is a live one on the composite path; it has nothing to do
with the repr, so this PR only records it.

## The debug line stops charging every layer

The nine eager `%`/f-string log calls in `composite/` take lazy arguments, so
the message is built only if a handler will take it. `_readable()` already did
this and stated why; most of the module did not. That is the cost the debug
line was charging on every layer, and it is why the raise was unconditional.

Scoped to `composite/` deliberately. `ruff --select G` finds 49 repo-wide; the
other 40 are in `psd/` read paths and three `api/` modules, none of which
interpolates a `Layer`. Adopting the `G` family is its own step under #692 and
#786, the way `D` was #772, so it is not ridden along here.

## Verification

- Full suite: 3028 passed, 3 skipped, 19 xfailed, 2 xpassed. ruff, ruff format
  and mypy clean.
- **Corpus sweep**: every fixture in `tests/psd_files`, per document *and* per
  layer, `force=False` and `force=True` — 4254 renders, bitwise identical to
  `main`. Every effect class present in the corpus (ten of them) is one the
  registry holds, so neither defect is reachable from a Photoshop file and both
  are forged in the tests.
- **Repr sweep**: every layer's `repr` across the corpus, 2127 of them,
  byte-identical to `main`.
- **A/B**: reverting each of the four parts separately fails at least one new
  test, and none passes for free. Narrowing the repr's clause to `_UNREADABLE`
  fails exactly one parametrized case.

## Found and left out

`Effects.scale` raises `ValueError("Effects data is None")` for any layer with
no effects block, which the second part above makes the normal state for a
layer whose block did not parse — so `scale` raises there where `enabled`,
`__len__`, `find()` and `__repr__` all answer cleanly. It is pre-existing, no
caller in the library reads it, and making it return its own `100.0` default
would be an API change with its own changelog line. Left as it is, and the
comment beside the guard names it rather than claiming it away.

`has_effects(enabled=False)` returns `True` whenever an effects block is
present, without asking whether anything in it can be listed — so it now says
`True` for a layer whose every effect was skipped, as it already did for the
38 corpus layers (of 117 with an effects tag) whose effects are all
not-present. Pre-existing, reachable from an ordinary Photoshop file, and
fixing it flips those 38: filed as #830 with the measurement rather than
ridden along here.

Two claims from an earlier draft of this work were checked and removed rather
than shipped: pytest's assertion rewriting does **not** lose the real failure
to a raising repr (it routes operands through `saferepr`), and a logging
handler does recover from a failed lazy format. Both were in a docstring
arguing for `except Exception`; the argument stands without them.


kyamagu:
Thanks — all three checked. Two were right and are fixed in 5357683; one is
real but predates this PR, so it is deferred to #830.

### Fixed: the DEBUG case never reached the handler

Correct, and worse than "does not invalidate the cache" suggests — the case
was passing vacuously. Counting records on the compositor's logger:

| | records emitted |
|---|---|
| DEBUG, cold cache | 2 |
| DEBUG, after the WARNING case ran | **0** |

So the parametrization asserted nothing about handler-side formatting. Now
`setLevel()`, restored through a finalizer, and the handler counts what it is
given so the test states which case must reach it and which must not:

```python
assert (records.count > 0) is (level == logging.DEBUG)
```

Reverting to `logger.level = level` fails the DEBUG case, so the fix is
pinned rather than assumed.

### Fixed: the changelog named the wrong exception

Correct. `ValueError` is only the unknown-class half; an effects block that
did not parse raised `IndexError` or `AttributeError` depending on its bytes —
which the PR body's own table shows, so the entry contradicted it. It names no
exception now.

### Deferred to #830: `has_effects(enabled=False)`

The inconsistency is real and stated accurately, but it is not introduced
here. Measured over `tests/psd_files` on `main` at 310bba05, before any of
this:

| layers with an effects tag | of which zero listable effects |
|---|---|
| 117 | 38 |

`has_effects(enabled=False)` already returns `True` for all 38. Photoshop
writes the effects block with a `present` flag per effect type, so a layer
that merely *had* an effect carries a block with nothing present in it — this
is reachable from an ordinary file, not from the malformed ones this PR is
about. What changed here is only that two more shapes route into that branch,
which is how you came to notice it.

Fixing it flips the answer for those 38 corpus layers: an `[api]` behaviour
change on ordinary files, and the question of *which* answer is wanted —
"is there an effects block" or "are there effects to list" — is a maintainer's
call, not an obvious bug. That deserves its own entry and its own decision
rather than riding along on a robustness fix, so it is #830 with the
measurement and both readings written out.

PR #832, opened 2026-09-17
fix: avoid seeding pen-only vector rasters

### LINKED ISSUE #823: composite: _draw_path() seeds a pen-only raster as fully covered
## Summary

`composite.vector._draw_path()` seeds its plane as fully covered when the
layer's vector mask has an initial fill rule and carries no paths:

```python
color = 0
if layer.vector_mask.initial_fill_rule and len(layer.vector_mask.paths) == 0:
    color = 1
mask = np.full((height, width, 1), color, dtype=np.float32)
```

That is right for `draw_vector_mask()`, where it means "a reveal-all mask with
nothing subtracted from it". It is applied regardless of whether the caller
passed a `brush` or a `pen`, so for `draw_stroke()` it means **the entire
raster is stroke** — a layer with no path to outline comes back fully covered
instead of empty.

## Reproduction

`adjustment-fillers.psd`'s `Color Fill 1` is the seed condition without a
stroke, so `_draw_path()` can be called on it directly:

```python
from psd_tools.composite import vector

psd = PSDImage.open("tests/psd_files/adjustment-fillers.psd")
layer = [x for x in psd.descendants() if x.name == "Color Fill 1"][0]
layer.vector_mask.initial_fill_rule, len(layer.vector_mask.paths)  # (1, 0)

vector._draw_path(layer, brush={"color": 255}).max()              # 1.0 -- correct
vector._draw_path(layer, pen={"color": 255, "width": 1.0}).min()  # 1.0 -- a stroke covering everything
```

## Reachability

Not reachable from the fixture corpus, and possibly not from Photoshop's UI
either. `_get_stroke()` needs `layer.stroke.enabled` and a vector mask, and the
two conditions do not co-occur today: six layers have the seed condition
(`adjustment-fillers.psd`, `masks/2.psd` x2, `vector-mask.psb`,
`vector-mask2.psd`, `vector-mask3.psd`) and none of them has a stroke, while
all 49 stroked layers in the corpus have at least one path.

So this is latent rather than a live rendering bug. It would take a shape layer
whose paths had been stripped — a malformed file, or one written by another
tool.

## Why file it now

#807 changed the blast radius. `_get_stroke()` used to rasterize at document
size and `paste()` the result onto the viewport, which clipped the all-ones
plane to the canvas. Since #822 the raster is made on the viewport directly, so
the plane now covers the whole viewport including anything off-canvas. Both are
wrong; the new one is wrong over a larger area.

## Suggested fix

Gate the seed on the brush, since it describes a fill rule and a pen has no
fill to rule on:

```python
if brush and layer.vector_mask.initial_fill_rule and len(layer.vector_mask.paths) == 0:
    color = 1
```

The same question applies to the two `if first and brush:` inversions further
down, which are already gated this way — the seed looks like the one that was
missed.

Found while reviewing #822.

dydydd:
## Summary

Gate the reveal-all fill-rule seed in `composite.vector._draw_path()` on the presence of a brush. Pen-only rendering has no fill rule, so a path-less pen must not produce a fully covered raster.

Fixes #823.

## Verification

- Pen-only seed contract: 6 passed
- Full test suite: 3034 passed, 3 skipped, 19 xfailed, 2 xpassed
- Composite tests: 1212 passed, 17 xfailed, 2 xpassed
- API tests: 413 passed
- Ruff check and format: passed
- Mypy: passed


kyamagu:
@dydydd Thanks!

